PR #1, opened 2023-06-08
SymCryptSha256 and SymCryptSha384 functions

nnmkhang:
This change includes the following:

SymCrypt-sys:
- build.rs script that produces bindings 
- symcrypt_bindings.rs which is the modified bindings produced by the bindgen. This is what the symcrypt crate will consume

SymCrypt:
- Friendly Rust wrappers for Sha384, Sha256 and related functions
- Simple main.rs program that showcases the Sha256 and Sha384 working correctly


symcrypt/src/lib.rs
```diff
@@ -0,0 +1,141 @@
...
+                is_dirty: true
+            };
+            unsafe {
+                symcrypt_sys::SymCryptSha256Init(&mut instance.state);
+            }
+        instance
+    }
+
+    pub fn append(&mut self, data: &[u8] ) {
+        unsafe {
```
PhilHallin: is_dirty should be set  #Closed

symcrypt/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+            symcrypt_sys::SymCryptSha256Append(
+                &mut self.state, // pState
+                data.as_ptr(), // pbData
+                data.len() as symcrypt_sys::SIZE_T, //cbData
+            );
+        }
+        self.is_dirty = true;
```
PhilHallin: >self.is_dirty = true; [](http://example.com/codeflow?start=8&length=21)

Did you run through any rust pretty formatter?

For instance, do you need the trailing ";" for the last statement? #Closed

symcrypt/src/lib.rs
```diff
@@ -0,0 +1,143 @@
...
+    }
+}
+
+impl Drop for Sha256State {
+    fn drop(&mut self) {
+        if self.is_dirty {
+            unsafe {
+                symcrypt_sys::SymCryptWipe( 
+                ptr::addr_of_mut!(self.state) as *mut c_void, 
+                mem::size_of_val(&mut self.state) as symcrypt_sys::SIZE_T)
```
PhilHallin: >mu [](http://example.com/codeflow?start=34&length=2)

Is the mut necessary? #Closed

PhilHallin:
> /* automatically generated by rust-bindgen 0.65.1 */

Is the same symcrypt_binding.rs file be generated for both Windows and Linux?

---
Refers to: symcrypt-sys/src/symcrypt_bindings.rs:1 in b539543. [](commit_id = b539543a3e1c453985a2b6e85d300aa4afe83ab0, deletion_comment = False)

PhilHallin:
> pub struct _SYMCRYPT_MD5_CHAINING_STATE {

Why are MD5 and SHA1 symbols being generated?

---
Refers to: symcrypt-sys/src/symcrypt_bindings.rs:50 in b539543. [](commit_id = b539543a3e1c453985a2b6e85d300aa4afe83ab0, deletion_comment = False)

symcrypt/src/lib.rs
nnmkhang: Did not run through a prettier formatter, I will see if I can include one in the pipeline for commits on github.

symcrypt/src/lib.rs
nnmkhang: Yes

nnmkhang:
@PhilHallin  in regards to the questions about the bindings being generated:

Q: Is the same symcrypt_binding.rs file be generated for both Windows and Linux?
A: No there is a slight difference between them, I have not ran into any issues when running any code so far though. For now we'll keep the binding files as one until there a differential is needed, at that point we can use either section parts off for linux/windows or create separate binding files.

Q: Why are MD5 and SHA1 symbols being generated?
A: This was generated recursively with bindgen, [LINK](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.allowlist_function), MD5 and SHA1 probably got included somewhere in bindgens recursive search. Going forward, we can clean up the bindgen file and tighten up the regex to hopefully catch less stray functions being produced.

PhilHallin (approved):
:shipit:

README.md
```diff
@@ -1 +1,4 @@
-# rust-symcrypt-sys
+this repo holds two crates; symcrypt (friendly rust wrappers), and symcrypt-sys (FFIs)
```
mlindgren, line 1: >symcrypt-sys [](http://example.com/codeflow?start=67&length=12)

I think this name could be confusing; to me it kind of implies that this is a kernel driver, which is not the case. Maybe call this `symcrypt-ffi`? #Resolved

symcrypt-sys/build.rs
```diff
@@ -0,0 +1,90 @@
...
+        // symcrypttestmodule* files must be placed in Windows/System32/ as a workaround at the moment.
+    }
+    
+    #[cfg(target_os = "linux")]
+    {   
+        println!("cargo:libdir=./inc");
+        println!("cargo:rustc-link-lib=dylib=symcrypt"); // the lib prefix for libsymcrypt is implied on linux
+
+        // TODO: Create a script that copies all libsymcrypt.so* files from SymCrypt path to /lib/x86_64-linux-gnu/
+        // The ld linker will look for the symcrypt.so files within /lib/x86_64-linux-gnu/. No need to set a hardcoded path.
```
mlindgren, line 22: These files will already be present on distros that ship with SymCrypt, such as Mariner. This would still be necessary on Ubuntu though, unless we create SymCrypt packages for distros like Ubuntu. That's not currently on the roadmap, though. #Resolved

symcrypt-sys/build.rs
```diff
@@ -0,0 +1,90 @@
...
+        .allowlist_function("SymCryptTlsPrf1_2ExpandKey")
+        .allowlist_function("SymCryptTlsPrf1_2Derive")
+        .allowlist_function("SymCryptTlsPrf1_2")
+
+        // HKDF functions 
+        .allowlist_function("^(SymCryptHkdf.*)$")
+
+        // ECDH Key Agreement
+        .allowlist_var("SymCryptEcurveParamsNistP256")
+        .allowlist_var("SymCryptEcurveParamsNistP384")
```
mlindgren, line 66: Were you able to resolve the linking issue with these exported constants? #Resolved

symcrypt/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    let data = b"this is a test";
+    let mut result: [u8; SHA256_RESULT_SIZE] = [0; SHA256_RESULT_SIZE];
+    sha256(data, &mut result);
+    println!("SHA256 stateless hash: {:?}", result);
+
+    let mut sha_test = Sha256State::new();
+
+    let mut result: [u8; SHA256_RESULT_SIZE] = [0; SHA256_RESULT_SIZE];
+    Sha256State::append(&mut sha_test, b"this is a test");
```
mlindgren: >b"this is a test") [](http://example.com/codeflow?start=39&length=18)

Use `data` here as well? #Resolved

symcrypt/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+    Sha256State::result(&mut sha_test, &mut result);
+    println!("SHA256 state hash: {:?}", result);
```
mlindgren, line 16: >println!("SHA256 state hash: {:?}", result); [](http://example.com/codeflow?start=4&length=44)

Suggest comparing the result against a known-correct value. Same for SHA 384.

symcrypt-sys/inc/precomp.h
```diff
@@ -0,0 +1,16 @@
+//
+// SymCrypt linux shared object library pre-compiled header file
+//
+// Copyright (c) Microsoft Corporation. Licensed under the MIT license.
+//
+
```
mlindgren, line 6: You should not be duplicating the SymCrypt headers in your repo. They're subject to change and will likely get out-of-sync quickly as we continue development on SymCrypt.

Instead, you should either get them from a package on the host system (e.g. the SymCrypt package on Mariner includes the headers), or make SymCrypt a submodule of this project. The former is preferable, but might not be practical right now since we don't have SymCrypt packages for many common Linux distros (like Ubuntu).

mlindgren:
> /* automatically generated by rust-bindgen 0.65.1 */

Should auto-generated files be checked in? Shouldn't this be generated every time as part of the build?

---
In reply to: [1589889429](https://github.com/nnmkhang/rust-symcrypt-sys/pull/1#issuecomment-1589889429) [](http://example.com/codeflow?ancestors=1589889429,1589889429)

---
Refers to: symcrypt-sys/src/symcrypt_bindings.rs:1 in b539543. [](commit_id = b539543a3e1c453985a2b6e85d300aa4afe83ab0, deletion_comment = False)

README.md
PhilHallin: -sys is the common naming convention for rust bindings to external C DLLs. For mappings to other windows APIs, see the windows-sys crate.

mlindgren: Ah, that's unfortunate because it does seem confusing to me from a Windows perspective, but if that's the Rust convention then I suppose we should follow that.

symcrypt/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+
+pub const SHA256_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA256_RESULT_SIZE as usize;
+pub const SHA384_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA384_RESULT_SIZE as usize;
+
+pub struct SymCryptInit;
+impl SymCryptInit {
+    pub fn new() {
+        unsafe {
+            //symcrypt_sys::SymCryptInit();
+            // TODO: Find out why SymCryptInit() breaks on linux / windows (BREAKING)
```
mlindgren, line 14: >// TODO: Find out why SymCryptInit() breaks on linux / windows (BREAKING) [](http://example.com/codeflow?start=12&length=73)

What do you mean by "breaks"? Is it crashing? What's the error?

symcrypt/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+        unsafe {
+            symcrypt_sys::SymCryptSha256Result(&mut self.state, result.as_mut_ptr());
+        }
+        self.is_dirty = false;
+    }
+}
+
+impl Drop for Sha256State {
+    fn drop(&mut self) {
+        if self.is_dirty {
```
mlindgren: I suggest getting rid of this variable and just unconditionally wiping when the state is destroyed/dropped/whatever the Rust terminology is. I don't think the marginal perf benefit is worth the risk of forgetting to set this value and not wiping the state when it should be wiped. #Resolved

symcrypt/src/lib.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub fn new() {
+        unsafe {
+            //symcrypt_sys::SymCryptInit();
+            // TODO: Find out why SymCryptInit() breaks on linux / windows (BREAKING)
+        }
+    }
+}
+
+pub struct Sha256State {
+    state: symcrypt_sys::_SYMCRYPT_SHA256_STATE,
```
mlindgren: >_SYMCRYPT_SHA256_STATE [](http://example.com/codeflow?start=25&length=22)

Nit: can you use the typedef'd name (without the preceding underscore)? It looks like bindgen generated an alias for it. #Resolved

mlindgren (changes requested):
:clock1:

symcrypt/src/lib.rs
PhilHallin: We are talking about clearing 100+ memory bytes.

In about all cases, result would have been previously call where it will do the appropriate wipe already.

I thought you crypto guys cared about performance???

This is the low level rust code that will be called for every hash operation.

mlindgren:
Drop is only called when the object is deallocated (goes out of scope),
right? If you're doing many hash operations, you can keep a single hash
object alive and reuse it, and then you'd only incur one extra wipe. To
me that feels worthwhile for the additional safety. (The code as-is
appears correct to me; I'm more concerned about potential future changes
that could forget to set this.) Samuel is more of an expert on perf,
though, so I'd like to hear his input as well.

On June 13, 2023, GitHub ***@***.***> wrote:
> @PhilHallin commented on this pull request.
>

> In symcrypt/src/lib.rs <https://github.com/nnmkhang/rust-symcrypt-
> sys/pull/1#discussion_r1228860089>:
>
> > + ); + } + self.is_dirty = true; + } + + pub fn result(&mut self,
> result: &mut [u8; SHA256_RESULT_SIZE]) { + unsafe { +
> symcrypt_sys::SymCryptSha256Result(&mut self.state,
> result.as_mut_ptr()); + } + self.is_dirty = false; + } +} + +impl Drop
> for Sha256State { + fn drop(&mut self) { + if self.is_dirty {
> We are talking about clearing 100+ memory bytes.
>
> In about all cases, result would have been previously call where it
> will do the appropriate wipe already.
>
> I thought you crypto guys cared about performance???
>
> This is the low level rust code that will be called for every hash
> operation.
>

—
Reply to this email directly, view it on GitHub
<https://github.com/nnmkhang/rust-symcrypt-
sys/pull/1#discussion_r1228860089>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAHGYDAF43PA4XB43NNB4ODXLEF3BANCNFSM6AAAAAAY72RG54>.
You are receiving this because you commented.Message ID: <nnmkhang/rust-
***@***.***>

symcrypt/src/lib.rs
mlindgren: Drop is only called when the object is deallocated (goes out of scope), right? If you're doing many hash operations, you can keep a single hash object alive and reuse it, and then you'd only incur one extra wipe. Also, if you don't keep a single object alive, you'll be reallocating for every hash so it will be slow anyway.

To me that feels worthwhile for the additional safety. (The code as-is appears correct to me; I'm more concerned about potential future changes that could forget to set this.) @samuel-lee-msft is more of an expert on perf, though, so I'd like to hear his input as well.

symcrypt-sys/build.rs
nnmkhang: Noted, I will keep that in mind when writing the instillation instructions later down the line

nnmkhang:
> > /* automatically generated by rust-bindgen 0.65.1 */
> 
> Should auto-generated files be checked in? Shouldn't this be generated every time as part of the build?
> 
> In reply to: [1589889429](https://github.com/nnmkhang/rust-symcrypt-sys/pull/1#issuecomment-1589889429)
> 
> Refers to: symcrypt-sys/src/symcrypt_bindings.rs:1 in [b539543](https://github.com/nnmkhang/rust-symcrypt-sys/commit/b539543a3e1c453985a2b6e85d300aa4afe83ab0). [](commit_id = [b539543](https://github.com/nnmkhang/rust-symcrypt-sys/commit/b539543a3e1c453985a2b6e85d300aa4afe83ab0), deletion_comment = False)


This is a somewhat common practice for bindgen, which is the tool we're using for the binding generation. Bindgen does a pretty good job but its not perfect. Checking in the generated and fixed bindings is preferable because it saves the user on build time, and allows for better correctness

symcrypt-sys/build.rs
nnmkhang: Yes, we were able to fix them with this:

`#[link(name = "symcrypttestmodule", kind = "dylib")]
extern "C" {
    pub static SymCryptSha256Algorithm: PCSYMCRYPT_HASH;
}`

Copied from internal email thread:
"
The core issue is something mentioned at the bottom of the docs for [EXPORTS | Microsoft Learn](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Fcpp%2Fbuild%2Freference%2Fexports%3Fview%3Dmsvc-170&data=05%7C01%7Ckhangnguyen%40microsoft.com%7Cc4649113c171424fe3c108db685f3dbb%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638218530497811201%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=UCMGw0CzGkk1T1TSif2g2tZI7yHaiNXjEjIAxxNyFbk%3D&reserved=0):
> When you export a variable from a DLL by using a .DEF file, you do not have to specify __declspec(dllexport) on the variable. However, in any file that uses the DLL, you must still use __declspec(dllimport) on the declaration of data.
(emphasis mine)
"

This requires change for the .def export files to append the DATA keyword on the SymCrypt Side.

symcrypt/src/main.rs
nnmkhang: Good Idea, is there a value that you suggest?

symcrypt/src/lib.rs
nnmkhang: There is a link error and the rest of the code will not compile and or run. We can debug offline regarding this one

symcrypt-sys/inc/precomp.h
nnmkhang: I agree, this is a workaround for prototyping.

> They're subject to change and will likely get out-of-sync quickly as we continue development on SymCrypt.

I do have a follow up questions;
Is SymCrypt backwards compatible with all updates? if that is the case then we can look at the idea of searching the system for the most up to date header files to include. 

At the moment, we are only taking a subsection of SymCryptAPIs, and we're doing some manual modifications to the binding files to ensure correctness ( DATA vs CODE export issue ). If we are already taking these precautions would it make sense to manually manage and update the SymCrypt bindings on a need to update basis? As an example; every major version SymCrypt update. My worry with getting the latest SymCrypt versions would cause breaking changes between versions.

> since we don't have SymCrypt packages for many common Linux distros (like Ubuntu).

Is there any plan in the future to include a pkg-config file to the SymCrypt repo for Linux users? From what I have seen this is the standard method for finding and linking libraries as well as linking in header files. Also what is the distribution story on Windows in the future? The plan is to have a SymCrypt.dll ship with Windows but where where would we get the corresponding header files?

symcrypt/src/lib.rs
nnmkhang: Will do

symcrypt/src/main.rs
nnmkhang: Fixed

symcrypt-sys/Cargo.lock
```diff
@@ -0,0 +1,364 @@
...
+
+[[package]]
+name = "shlex"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43b2853a4d09f215c24cc5489c992ce46052d359b5109343cbafbf26bc62f8a3"
+
+[[package]]
+name = "symcrypt-sys"
+version = "0.1.0"
```
samuel-lee-msft, line 234: I wonder if the version here should reflect the underlying SymCrypt version?

symcrypt/src/lib.rs
PhilHallin: Is SymCryptInit being exported from the DLL? The last we looked I didn't think it was exported??

symcrypt-sys/Cargo.lock
nnmkhang: Im not sure about rust naming best practice's, but that seems reasonable to me, we should table the naming and come back to it once we have a better understanding of future packaging. see [here](https://github.com/nnmkhang/rust-symcrypt-sys/pull/1#discussion_r1228829303)

symcrypt/src/lib.rs
samuel-lee-msft: I'm in two minds here - I agree with Mitch that the performant path is reusing a Sha256State; if you care about the performance of `drop` then you also care about the performance of `new`, and in `new` it's unavoidable that you're calling `symcrypt_sys::SYMCRYPT_SHA256_STATE::default()` (itself wiping state using Rust's write_bytes), and `SymCryptSha256Init` redundantly vs. reusing the state.

However, it does seem fairly simple to keep track of `is_dirty` correctly here, and it would remove a SymCryptWipe in most programs using Sha256State.

I would be minded to keep it simple until there is a clear need for optimization.

PhilHallin: Most applications won't be re-using this object after calling result(). However, they will be making multiple calls to append(). It they only need one append they would call the stateless sha256() function.

Note, I thought this was simple??  For sha256, 128 bytes to be wiped. For sha384, 224 bytes.

symcrypt/src/main.rs
samuel-lee-msft: If this is a test program, it might be worth ingesting some other Rust crypto library and comparing outputs.

You can also hardcode the output you have got from the test program so far back into the test program, and double check with an online hash generator / python / whatever.
i.e. I believe the result with SHA256 for your current data should be: 2e99758548972a8e8822ad47fa1017ff72f06f3ff6a016851f45c398732bc50c

symcrypt/src/lib.rs
samuel-lee-msft: "Most applications won't be re-using this object after calling result()." Then most applications will not be very performant because they will be calling `new` repeatedly; they shouldn't care about the extra 10s of cycles to Wipe the state redundantly.

It is simple with the `is_dirty` flag in this specific context, I think the question is how much logic we want to have in this layer generally, and I think to start with we should minimize this. Definitely can revisit if there's a clear performance issue to handle.
It seems a bit wonky to have extra state/logic in this Rust layer; if we need to have performant single-use stateful hashing perhaps we should be exposing a `SymCryptShaXxxCleanup` function which wipes whatever is necessary.

PhilHallin: OK we will remove this is_dirty optimization. Won't consider any other optimizations in this layer. If necessary can be added in the future without changing the rust application contract. Thanks.

symcrypt/src/main.rs
nnmkhang: I have checked the test outputs with the online generators and can confirm that it is the same. I will eventually move this testing functionality over to a test file where we can assert against the hardcoded known values ex: 2e99758548972a8e8822ad47fa1017ff72f06f3ff6a016851f45c398732bc50c

symcrypt/src/lib.rs
nnmkhang: Sounds good, I will remove the is_dirty

symcrypt-sys/inc/precomp.h
nnmkhang: Resolving, we can table this and leave it as an ongoing discussion since it does not block the progression of other wrappers. Also made an issue to track.

symcrypt/src/lib.rs
mlindgren: Oh yeah, that's a good point. SymCryptInit is for statically linked libraries. With the DLL, you should use SymCryptModuleInit instead.

symcrypt-sys/inc/precomp.h
mlindgren: We use semantic versioning, meaning that we guarantee compatibility within a given major version. We're currently on major version 103. If there are breaking changes in the future, we will bump that to 104. Until then, future versions of 103.x will be compatible with existing versions.

If you add SymCrypt as a submodule of your repo, it will be pinned to the specific commit you choose. It won't change out from under you; you have to manually update the submodule to pull in new changes. I think this is a good solution until we have a more formal packaging system you can rely on (e.g. vcpkg or using the distro's package manager).

We do have a pkg-config file in the repo as well.

Please don't put this off for a future change. Duplicating files is asking for trouble; it allows the code to be forked and then it becomes difficult to undo that.

samuel-lee-msft:
Approved with the proviso that we should take out the symcrypt headers ASAP, and use a git submodule instead.

samuel-lee-msft (approved):

samuel-lee-msft (approved):

PhilHallin (approved):

PR #9, opened 2023-07-05
Adding HMAC code for Sha384 and Sha256

nnmkhang:
Added:
- HMAC Sha256
- HMAC Sha384
- dev dependency on hex crate

Changed:
- changed lib.rs to take in modules
- moved hash functions to a hash.rs module
- fixed hashing tests to use hex crate
- fixed SymCryptInit() by changing to SymCryptInitModule()


rust-symcrypt/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "symcrypt"
+name = "rust-symcrypt"
```
eholk, line 2: What's the reason for this change? The [Rust API guidelines](https://github.com/rust-lang/api-guidelines/blob/master/src/naming.md#casing-conforms-to-rfc-430-c-case) recommend against using `rust-` as a prefix, although obviously many packages do so.

rust-symcrypt/src/main.rs
```diff
@@ -1,31 +1,6 @@
...
-    let mut sha_test = Sha384State::new();
-
-    let mut result: [u8; SHA384_RESULT_SIZE] = [0; SHA384_RESULT_SIZE];
-    Sha384State::append(&mut sha_test, data);
-    Sha384State::result(&mut sha_test, &mut result);
-    println!("SHA384 state hash: {:?}", result);
-}
+use rust_symcrypt::SymCryptInit;
+
+fn main() {
```
eholk: It seems like this is meant to be more of an example program, right? If so, I'd recommend moving it to an `examples/` directory. See https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples for more information.

Having a main file here is meant for something where you are shipping both a library (for example, so that other Rust programs can use symcrypt) as well as a binary (such as a command line tool that can sign and verify files). I was under the impression that this crate is mainly meant to be a symcrypt library without including a standalone tool.

rust-symcrypt/src/main.rs
```diff
@@ -1,31 +1,6 @@
...
+    println!("Hello World :)");
+    SymCryptInit::new();
```
eholk: This feels a little weird as is, but maybe that's just because it's incomplete. `SymCryptInit::new()` looks like we're creating some `SymCryptInit` object and then immediately dropping it, where this looks more to me like it should be just a function call, like `symcrypt::init()`, where `init` is a function exported from the `symcrypt` module.

But, if in normal operation you'll save the `SymCryptInit` into a variable and either pass it to other functions or call methods on it, then the way you have it here makes sense.

rust-symcrypt/src/hash.rs
```diff
@@ -0,0 +1,186 @@
...
+    state: symcrypt_sys::SYMCRYPT_SHA256_STATE,
+}
+
+impl Sha256State {
+    pub fn new() -> Self {
+        let mut instance = Sha256State {
+            state: symcrypt_sys::SYMCRYPT_SHA256_STATE::default(),
+        };
+        unsafe {
+            symcrypt_sys::SymCryptSha256Init(&mut instance.state);
```
eholk, line 26: It's generally good practice to add a `// SAFETY: ` comment on this `unsafe` block or function call that explains why this operation is safe. In this case it's probably just that unsafe is needed for FFI, but safety comments also commonly say things like "this pointer was verified to be correctly allocated and initialized in function X" or whatever makes sense.

eholk: This advice applies to the other unsafe blocks in this PR too, but I won't comment on each one individually.

rust-symcrypt/src/hash.rs
```diff
@@ -0,0 +1,186 @@
...
+        }
+        instance
+    }
+
+    pub fn append(&mut self, data: &[u8]) {
+        unsafe {
+            symcrypt_sys::SymCryptSha256Append(
+                &mut self.state,                    // pState
```
eholk: If you use rust-analyzer, does the `pState: ` inlay show up? If so, it's probably okay to leave off the `// pState` part, although that's a project style question.

rust-symcrypt/src/hash.rs
```diff
@@ -0,0 +1,186 @@
...
+        unsafe {
+            symcrypt_sys::SymCryptWipe(
+                ptr::addr_of_mut!(self.state) as *mut c_void,
+                mem::size_of_val(&mut self.state) as symcrypt_sys::SIZE_T,
+            )
+        }
+    }
+}
+
+pub fn sha256(data: &[u8], result: &mut [u8; SHA256_RESULT_SIZE]) {
```
eholk: Is this basically a convenience method for something like this?

```rust
let mut state = Sha256State::new();
state.append(data);
state.result(result);
``` #Closed

Comment (eholk) [inline rust-symcrypt/src/hash.rs]: I'd consider having this function return `[u8; SHA256_RESULT_SIZE]` instead of taking it as an out parameter. Out parameters are somewhat uncommon in Rust, especially because you can use `Result` to both return a value and sucess/failure information (speaking of which, can `SymCryptSha256` fail?).

I think that depends on how people will use this though. There could be a cost from extra memory copies and such, but the optimizer is usually pretty good at eliminating those, I think.

File: rust-symcrypt/src/hash.rs [hunk 7]
```diff
@@ -0,0 +1,186 @@
...
+        let mut instance = Sha384State {
+            state: symcrypt_sys::SYMCRYPT_SHA384_STATE::default(),
+        };
+        unsafe {
+            symcrypt_sys::SymCryptSha384Init(&mut instance.state);
+        }
+        instance
+    }
+
+    pub fn append(&mut self, data: &[u8]) {
```
eholk: Since this has basically the same methods as `Sha256State`, it might be worth pulling them into a trait like `HashState`. You could use associated types on that to account for the different result types.

rust-symcrypt/src/hash.rs
```diff
@@ -0,0 +1,186 @@
...
+        Sha384State::result(&mut sha_test, &mut result);
+        assert_eq!(hex::encode(result), expected);
+    }
+
+    #[test]
+    fn check_state_size() {
+        unsafe {
+            assert_eq!(
+                check_hash_size(SymCryptSha256Algorithm),
+                size_of::<Sha256State>() as u64
```
eholk: You can add `#[repr(transparent)]` on `Sha256State` and `Sha384State` if you want to provide a stronger guarantee that the sizes match.

rust-symcrypt/src/hash.rs
```diff
@@ -0,0 +1,186 @@
...
+    unsafe {
+        symcrypt_sys::SymCryptSha384(
+            data.as_ptr(),                      // pbData
+            data.len() as symcrypt_sys::SIZE_T, // cbData
+            result.as_mut_ptr(),                // pbResult
+        );
+    }
+}
+
+#[cfg(test)]
```
eholk, line 138: I like the tests you've added!

symcrypt-sys/src/hash.rs
eholk, line 1: I'm not sure you'll need this file here, since it looks like symcrypt-sys will be basically entirely auto-generated.

rust-symcrypt/src/hash.rs
nnmkhang: I use rust analyzer which does show the pState: But I wanted to keep the comments for clarity, but if majority of users have some sort of rust analyzer extension I will remove the // pState comments.

rust-symcrypt/Cargo.toml
nnmkhang: I originally had it named just "symcrypt" but I added the SymCrypt github as a subrepo so there was a conflict, I named the new crate "rust-symcrypt" because its what came up in my mind first, I will think about a different name here.

rust-symcrypt/src/main.rs
nnmkhang: I can remove this main.rs file, I was mainly keeping it to test things here and there but its not needed now since I've written unit tests.

symcrypt-sys/src/hash.rs
nnmkhang: Nice find, this one slipped by me. Will remove

rust-symcrypt/src/hash.rs
nnmkhang: Thanks!

rust-symcrypt/src/main.rs
nnmkhang: Thanks for this, I will change it over to be called as a regular function (`symcrypt::init()`)

rust-symcrypt/Cargo.toml
eholk: One option is to leave the repo name as rust-symcrypt but publish it to crates.io as symcrypt.

rust-symcrypt/src/hash.rs
eholk: As far as I know, the vast majority of Rust users use rust-analyzer. All the ones I know personally do, but I don't have any kind of representative survey data to say one way or the other.

rust-symcrypt/src/hash.rs
PhilHallin: I'm confused. All this module does is wrap FFI calls. Seems like adding SAFETY is redundant information.

rust-symcrypt/src/hash.rs
PhilHallin: Since this is the low level wrapper, we intentionally wanted to avoid any potential extra allocations. This was the recommendation from the symcrypt experts.

PhilHallin: It makes one symcrypt call where symcrypt can optimize.

So if you only need to hash one block, this would be the function to call.

eholk: `[u8; SHA256_RESULT_SIZE]` is a value type, so there's no heap allocation to worry about. But, it does get allocated on the stack and that can result in lots of extra memcpy traffic if the optimizer can't figure it out, which seems like the bigger concern to me here.

PhilHallin: None of these crypto functions can fail.

We want to avoid potential extra copies. In some crypto patterns there might be a collection of nested hashes to be hashed. That's why we wanted the caller to provide the slice to be updated.

rust-symcrypt/src/hmac.rs
```diff
@@ -0,0 +1,215 @@
...
+            state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE::default(),
+            expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default(),
+        };
+        unsafe {
+            symcrypt_sys::SymCryptHmacSha256ExpandKey(
+                &mut instance.expanded_key,        // pExpandedKey
+                key.as_ptr(),                      // pbKey
+                key.len() as symcrypt_sys::SIZE_T, // cbKey
+            );
+            symcrypt_sys::SymCryptHmacSha256Init(&mut instance.state, &mut instance.expanded_key);
```
PhilHallin: >mut [](http://example.com/codeflow?start=71&length=3)

Does this need to be mut? Its an in read only parameter.

rust-symcrypt/src/hash.rs
PhilHallin: After thinking further, I agree with Eric on not using uncommon out parameters. Instead should be the function's return value.

rust-symcrypt/src/hash.rs
nnmkhang: I feel that it would be pretty common. I will remove the comments

rust-symcrypt/src/hash.rs
nnmkhang: Okay I can change it over, @mlindgren thoughts?

rust-symcrypt/src/hmac.rs
nnmkhang: Changed to non mut for expanded_key

rust-symcrypt/Cargo.toml
nnmkhang: Okay I will do that

rust-symcrypt/src/hash.rs
nnmkhang: Fixed via @eholk's commit

rust-symcrypt/src/hash.rs
nnmkhang: Fixed via @eholk's commit

rust-symcrypt/src/hash.rs
eholk: In this case it doesn't really add much extra information, but I think it's still good practice to do. Unsafe blocks without `SAFETY` comments always look a little suspicious to me while something like `// SAFETY: unsafe needed for FFI` let's me know that nothing fancy is meant to be going on inside that block.

rust-symcrypt/src/hash.rs
eholk: > We want to avoid potential extra copies. In some crypto patterns there might be a collection of nested hashes to be hashed. That's why we wanted the caller to provide the slice to be updated.

What does this pattern look like in pseudo code? It might be worth adding both options to provide both ease of use and control when needed.

rust-symcrypt/src/hash.rs
nnmkhang: I read over some documentation on `#[repr(transparent)]`, from what I understand it enforces the Rust struct has the same memory layout and ABI as the corresponding foreign struct. When Phil and I stepped through the debugger that already seemed to be the case. Do you know if there would be any added benefit for adding the `#[repr(transparent)]`? I think I am more inclined to just trust the rust compiler

rust-symcrypt/src/hash.rs
```diff
@@ -62,14 +62,16 @@ impl Drop for Sha256State {
     }
 }
 
-pub fn sha256(data: &[u8], result: &mut [u8; SHA256_RESULT_SIZE]) {
+pub fn sha256(data: &[u8]) -> [u8; SHA256_RESULT_SIZE] {
```
nnmkhang, line 65: Choosing [u8; SHA256_RESULT_SIZE] over vec because size is known. Would there be any drawbacks from this? I understand that vec is more common and easier to work with in Rust.

nnmkhang:
Note: I have moved creating the Hmac trait to another branch because I'm having some AV issues and don't want to block this one from getting merged.

rust-symcrypt/src/hash.rs
```diff
@@ -0,0 +1,208 @@
...
+            state: symcrypt_sys::SYMCRYPT_SHA256_STATE::default(),
+        };
+        unsafe {
+            symcrypt_sys::SymCryptSha256Init(&mut instance.state);
+        }
+        instance
+    }
+}
+
+impl Hash for Sha256State {
```
mlindgren, line 32: I'm not very familiar with macros in Rust; I wonder if you can/should use macros to implement these different hash traits since they're all the same except for the result size and function call names. We use macros for this in SymCrypt itself, although the readability/convenience tradeoff is debatable.

rust-symcrypt/src/hash.rs
mlindgren: Maybe I'm missing something, but it seems to me that we don't actually care if the memory layout, packing, etc of the rust `Sha*State` objects is the same as the SymCrypt ABI, because you are not passing those objects directly to SymCrypt. Rather, you pass the underlying `state` member. **That** member **must** have be exactly the same as the SymCrypt ABI or bad things will happen. I assume bindgen already guarantees that? Does it mark `symcrypt_sys::SYMCRYPT_SHA256_STATE` as `#[repr(transparent)]`?

rust-symcrypt/Cargo.toml
mlindgren: You can set a custom path for the submodule so that it doesn't conflict with the name you want to use. Long-term, we want to move to vcpkg anyway, so I think that's the right solution.

rust-symcrypt/src/hmac.rs
```diff
@@ -0,0 +1,215 @@
+use core::ffi::c_void;
+use std::mem;
+use std::ptr;
+use symcrypt_sys;
+
+pub const SHA256_HMAC_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA256_RESULT_SIZE as usize;
+pub const SHA384_HMAC_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA384_RESULT_SIZE as usize;
+
+pub struct HmacSha256State {
```
mlindgren, line 9: Same comment here regarding whether macros could/should be used here to reduce repetition in these implementations.

rust-symcrypt/src/hmac.rs
```diff
@@ -0,0 +1,215 @@
...
+
+            symcrypt_sys::SymCryptWipe(
+                ptr::addr_of_mut!(self.expanded_key) as *mut c_void,
+                mem::size_of_val(&mut self.expanded_key) as symcrypt_sys::SIZE_T,
+            );
+        }
+    }
+}
+
+pub fn hmac256(key: &[u8], data: &[u8], result: &mut [u8; SHA256_HMAC_RESULT_SIZE]) {
```
mlindgren:
```suggestion
pub fn hmacsha256(key: &[u8], data: &[u8], result: &mut [u8; SHA256_HMAC_RESULT_SIZE]) {
```

Include SHA in the name; HMAC is a just a type of MAC, and the 256 is the size of the SHA2 digest. So `hmac256` is insufficiently clear.

rust-symcrypt/src/hmac.rs
```diff
@@ -0,0 +1,215 @@
...
+
+            symcrypt_sys::SymCryptWipe(
+                ptr::addr_of_mut!(self.expanded_key) as *mut c_void,
+                mem::size_of_val(&mut self.expanded_key) as symcrypt_sys::SIZE_T,
+            );
+        }
+    }
+}
+
+pub fn hmac384(key: &[u8], data: &[u8], result: &mut [u8; SHA384_HMAC_RESULT_SIZE]) {
```
mlindgren:
```suggestion
pub fn hmacsha384(key: &[u8], data: &[u8], result: &mut [u8; SHA384_HMAC_RESULT_SIZE]) {
```

rust-symcrypt/src/hash.rs
mlindgren: Personally, I'd be inclined to go with the idiomatic solution of returning an array for now. Before we do any sort of public release, we should benchmark the wrappers and see what kind of overhead they're adding. If we find that Rust's optimizer is not smart enough to remove memcpy and this causes a perf problem, then we can change the API.

mlindgren (changes requested):
Looks good to me except for the names of the stateless HMAC functions. Besides those, other comments are suggestions/food for thought.

rust-symcrypt/src/hash.rs
eholk: In my opinion, `[u8; SHA256_RESULT_SIZE]` is better than `Vec<u8>` in pretty much every way given that you know the size is always going to be exactly `SHA256_RESULT_SIZE`. The most important difference is that `Vec` always involves a heap allocation, while the fixed sized array can be stored inline. If you want to put store the result on the heap, you can do `Box::new(sha256(data))`, which gives you something similar to but not quite a `Vec`. 

Another option to consider is making the result something like `Sha256Result`, defined like this:

```rust
struct Sha256Result {
    data: [u8; SHA256_SIZE],
}
```

Basically, the hides the fact that the hash is an array of bytes, which may or may not be useful. It depends on whether you want people to manipulate the bytes directly, or if the hash should always be treated semantically as a single unit?

rust-symcrypt/src/hash.rs
eholk: This is a pretty common use of macros in Rust too. A lot of the standard library is implemented this way, such as all the methods on `u8`, `u16`, `u32`, `u64`, and so on.

rust-symcrypt/src/hash.rs
nnmkhang: Noted, I will return an array and make an issue to track benchmarking the wrappers.

rust-symcrypt/src/hash.rs
nnmkhang: Bindgen doesn't  add `#[repr(transparent)]` but it does provide a self test which taking a look should be sufficient in testing the layout is the same. [bindgen_test_layout__SYMCRYPT_SHA256_STATE](https://github.com/nnmkhang/rust-symcrypt-sys/blob/main/symcrypt-sys/src/symcrypt_bindings.rs#L158)

rust-symcrypt/src/hash.rs
nnmkhang: I think returning the array would be easier because it gives the user an option if needed to modify/splice easily, it is also what is returned in symcrypt. 

On the other hand the function that I'm working on right now; `SymCryptCGM*` can fail and will need to have a struct to wrap the symcrypt errors/success. If we're going to implement a return type for the `SymCryptGCM*`, does it make sense to have wrappers for the other functions that don't necessarily need it? ( Hmac, Hash )

rust-symcrypt/src/hash.rs
nnmkhang: Removed code since the bindgen tests cover the size and layout tests.

rust-symcrypt/src/hash.rs
eholk: So for functions that can fail, the normal thing to do is return [`Result`](https://doc.rust-lang.org/std/result/). The `Ok` part is just whatever type the function would return on success. For FFI crates like this, the `Err` part usually wraps some enum that lists all the errors that could be returned by SymCrypt. It's good practice to use [thiserror](https://crates.io/crates/thiserror) to `#[derive(Debug)]` on your error type as well because that will make it play better with the rest of Rust's error handling.

PhilHallin (approved):
:shipit:

Issue #10, opened 2023-07-05
Discuss moving code over to Traits 


nnmkhang:
Rust's trait feature can be leveraged to make the code more rust idiomatic.

PR #11, opened 2023-07-06
Experiment with a trait for Hash functions

eholk:
See #10.

This shows an example of what a `Hash` trait might look like. I also wrote a generic test function that shows how you can write code that is largely agnostic about what hash function it's using.

I think this can still be improved quite a bit, but I wanted to share the idea so far.

Feel free not to merge this, I'm mostly posting it as an example.

### DIFF
--- a/rust-symcrypt/src/hash.rs
+++ b/rust-symcrypt/src/hash.rs
@@ -10,7 +10,7 @@ pub trait Hash {
     type Result;
 
     fn append(&mut self, data: &[u8]);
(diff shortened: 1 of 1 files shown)


eholk:
I pushed a new version which uses a return type instead of taking an out param. It makes the code someone cleaner (in my opinion) but might come with some performance differences. I'm posting it mostly as an example.

Each of the commits in this PR should basically stand on their own.

PR #16, opened 2023-07-20
Add GCM and ChaCha Poly encrypt/decrypt

nnmkhang:
Added:
- GCM Encrypt/Decrypt
- ChaCha Poly Encrypt/Decrypt
- errors.rs file to make friendly enum for SymCrypt Errors

Modified:
- lib.rs: added errors, gcm and chacha as modules


rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+        ) {
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(dst),
+            err => Err(err.into()),
+        }
+    }
+}
+
+// SymCrypt requires pointer address to stay static through the scope, passing expand_key as an out
+// parameter to maintain static pointer address
+fn expand_key(key: &[u8], expanded_key: &mut symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY) -> Result<(), SymCryptError> {
```
nnmkhang: Opted for out parameter rather than `Box<>` either way the code is internal so not a big deal but the out parameter is slightly more clear than box for this circumstance in my opinion.

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+use crate::errors::SymCryptError;
+
+pub struct GcmState {
+    expanded_key: symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY,
+    state: symcrypt_sys::SYMCRYPT_GCM_STATE,
+}
+
+impl GcmState {
+    pub fn new(key: &[u8], nonce: &[u8]) -> Result<Box<Self>, SymCryptError> {
+        let mut instance = Box::new(GcmState {
```
nnmkhang: Have to `Box<>` since memory address for Self is moved around when returning. My guess is that it has to do with `Result<>` wrapping the returned values. When I did this code without a `Result<>` the code worked fine. To be safe, we should go back and modify other symcrypt structs to return a `Box<>`'d version. Created an issue to track this. #Closed

nnmkhang: The reason that memory address cannot change is because of underlying SymCrypt code. SymCrypt's magic checks require that function has the exact pointer. [Example](https://github.com/microsoft/SymCrypt/blob/main/lib/gcm.c#L334C1-L335C1)

rust-symcrypt/src/chacha.rs
```diff
@@ -0,0 +1,108 @@
+use std::vec;
+use symcrypt_sys;
+use crate::errors::SymCryptError;
+
+pub fn chacha20_poly1305_encrypt(
+    key: &[u8],
+    nonce: &[u8],
+    auth_data: Option<&[u8]>,
+    src: &[u8],
+    tag: &mut [u8; 16] // this is an out parameter.
```
nnmkhang: Created as an out parameter to match GCM functions

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+                &mut self.state,
+                data.as_ptr(),
+                dst.as_mut_ptr(),
+                dst.len() as symcrypt_sys::SIZE_T,
+            );
+            dst
+        }
+    }
+
+    pub fn encrypt_final(&mut self, tag: &mut [u8]) {
```
nnmkhang, line 94: Size of `tag` is not known at compile time, taking it as an out parameter so caller can specify. The alternative is to take in a tag length. I'm okay with doing either, having the out parameter seems better since it makes both Encrypt/Decrypt have the same call signature #Closed

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+}
+
+pub fn gcm_encrypt(
+    key: &[u8],
+    nonce: &[u8],
+    auth_data: Option<&[u8]>,
+    src: &[u8],
+    tag: &mut [u8], // This is an out parameter
```
nnmkhang: See above comment on `encrypt_final()`  #Closed

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    #[test]
+    fn test_validate_parameters() {
+        let nonce = hex::decode("cafebabefacedbaddecaf888").unwrap();
+        let auth_data = hex::decode("feedfacedeadbeeffeedfacedeadbeefabaddad2").unwrap();
+        let expected_tag = hex::decode("5bc94fbc3221a5db94fae95ae7121a47").unwrap();
+        let pt = hex::decode("d9313225f88406e5a55909c5aff5269a86a7a9531534f7da2e4c303d8a318a721c3c0c95956809532fcf0e2449a6b525b16aedf5aa0de657ba637b39").unwrap();
+
+        unsafe {
```
nnmkhang: Still thinking of how to wrap this so that is friendly, if you have any ideas please let me know #Closed

rust-symcrypt/src/gcm.rs
PhilHallin: Please add your comments here  to the code.

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+}
+
+// SymCrypt requires pointer address to stay static through the scope, passing expand_key as an out
+// parameter to maintain static pointer address
+fn expand_key(key: &[u8], expanded_key: &mut symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY) -> Result<(), SymCryptError> {
+    unsafe {
+        match symcrypt_sys::SymCryptGcmExpandKey(
+            expanded_key,
+            symcrypt_sys::SymCryptAesBlockCipher,
```
PhilHallin: > symcrypt_sys::SymCryptAesBlockCipher, [](http://example.com/codeflow?start=11&length=38)

This should use a rust enum type. BlockCipherType. #Closed

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+}
+
+impl GcmState {
+    pub fn new(key: &[u8], nonce: &[u8]) -> Result<Box<Self>, SymCryptError> {
+        let mut instance = Box::new(GcmState {
+            state: symcrypt_sys::SYMCRYPT_GCM_STATE::default(),
+            expanded_key: symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY::default()
+        });
+        unsafe {
+            expand_key(key, &mut instance.expanded_key)?;
```
PhilHallin: >expand_key(key, &mut instance.expanded_key)?; [](http://example.com/codeflow?start=12&length=45)

Does this need to be unsafe?

rust-symcrypt/src/gcm.rs
PhilHallin: After thinking further, recommend changing this to tag_length parameter and returning the tag in the Result.

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+            symcrypt_sys::SymCryptWipe(
+                ptr::addr_of_mut!(self.expanded_key) as *mut c_void,
+                mem::size_of_val(&mut self.expanded_key) as symcrypt_sys::SIZE_T,
+            );
+        }
+    }
+}
+
+pub fn validate_gcm_parameters(
+    block_cipher: symcrypt_sys::PCSYMCRYPT_BLOCKCIPHER,
```
PhilHallin: >  block_cipher: symcrypt_sys::PCSYMCRYPT_BLOCKCIPHER, [](http://example.com/codeflow?start=2&length=53)

This should be a rust enum. No callers should need to know about any symcrypt_sys information. #Closed

rust-symcrypt/src/gcm.rs
PhilHallin: After further thinking, recommend tag_length parameter and return tag in the return result.

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,377 @@
...
+        expand_key(key, &mut expanded_key)?;
+
+        let (auth_data_ptr, auth_data_len) = auth_data.map_or_else(
+            || (std::ptr::null(), 0 as symcrypt_sys::SIZE_T),
+            |data| (data.as_ptr(), data.len() as symcrypt_sys::SIZE_T),
+        );
+
+        let mut dst = vec![0u8; src.len()];
+
+        symcrypt_sys::SymCryptGcmEncrypt(
```
PhilHallin, line 197: > symcrypt_sys::SymCryptGcmEncrypt( [](http://example.com/codeflow?start=7&length=34)

Can you restrict unsafe to only the function call? #Closed

rust-symcrypt/src/chacha.rs
```diff
@@ -0,0 +1,108 @@
...
+        }
+    }
+}
+
+pub fn chacha20_poly1305_decrypt(
+    key: &[u8],
+    nonce: &[u8],
+    auth_data: Option<&[u8]>,
+    src: &[u8],
+    tag: &[u8]
```
PhilHallin: > &[u8] [](http://example.com/codeflow?start=8&length=6)

Shouldn't this be &[u8; 16]

rust-symcrypt/src/chacha.rs
PhilHallin: Since the tag is fixed length, you can always return in the Result as [u8; 16]

rust-symcrypt/src/gcm.rs
nnmkhang: What changed your mind?

PhilHallin: So far we haven't had any out parameters. Only out via the Result. Seems better to be consistent throughout.

rust-symcrypt/src/gcm.rs
nnmkhang: Added

rust-symcrypt/src/gcm.rs
nnmkhang: moved outside of unsafe

rust-symcrypt/src/gcm.rs
nnmkhang: done

rust-symcrypt/src/chacha.rs
nnmkhang: changed

rust-symcrypt/src/chacha.rs
nnmkhang: I can do that, want to hear thoughts from @eholk.

PhilHallin: Whats the debate? It must be exactly 16 bytes.

rust-symcrypt/src/gcm.rs
PhilHallin: You need to add a rust enum type that is passed to validate_gcm_parameters.

rust-symcrypt/src/chacha.rs
nnmkhang: debate to return the tag or have it as an out parameter

eholk: I assume this returns a vec with the encrypted version of `src`, right? What is the tag used for? (Speaking of which, doc comments would help make this a little clearer :) )

I'd probably steer you towards returning the tag rather than having it as an out parameter, especially since this returns a `Result`. That makes it harder to run into mistakes where, for example, you only cared about the `tag` and then you forgot to check whether the function succeeded.

It's a little tricky here because you're returning two values, but tuples work well for that (`Result<(Vec<u8>, [u8; 16]), SymCryptError>`) or even making a one off struct so you can name the fields and make it clearer what each one means.

rust-symcrypt/src/chacha.rs
eholk: I'm still seeing `&[u8]`.

rust-symcrypt/src/errors.rs
```diff
@@ -0,0 +1,58 @@
+use symcrypt_sys;
+use std::convert::From;
+
+#[derive(Debug)]
+#[derive(PartialEq)]
+pub enum SymCryptError {
```
eholk: It's probably worth adding `#[non_exhaustive]` to this enum. That will require people to have a catch all branch if they match on it, which means you can add new variants in the future.

rust-symcrypt/src/gcm.rs
eholk: This is something I chatted with @nnmkhang a bit offline about, and doing this perfectly right is going to be a little tricky. Returning a `Box` will work most of the time, but people will still be able get into trouble if they try hard enough. For example, they could do:

```rust 
let state: Box<GcmState> = GcmState::new()?;
let new_state = Box::new(*state);
```

This moves the previously created state into a new location in memory, which means the magic value will be wrong again.

To do this fully correctly, you'd need to use [`Pin`](https://doc.rust-lang.org/std/pin/index.html). Unfortunately, `Pin` is really hard to use.

I think using `Box` is reasonable for the sake of ease of use. As I understand it, we don't get undefined behavior if the magic is wrong, but the functions that check the magic will either fail or shutdown the program. If that's the case, I think it'd be fine to note that in the doc comments and maybe provide a function that updates the magic that you can call if you really need to move out of a box for some reason.

If you wanted to go the `Pin` route, I'd recommend trying to keep that out of the public API as much as possible. One way you could do that is to build your own wrapper types that handle pinning internally. I'm not 100% sure what this would look like though, so it'd need a little more design work. This might be a good idea even for boxes though, because you could prevent some of the cases where people can get into trouble, like the example I gave.

rust-symcrypt/src/gcm.rs
eholk: One advantage of the out parameter is that it lets the caller control memory allocation, while if we take a length and return the tag, that means `encrypt_final` will have to handle memory allocation (probably by returning a `Vec<u8>`). Some applications are very sensitive to memory allocation, so that extra control can be important, but it depends on the requirements you expect your users to have.

rust-symcrypt/src/gcm.rs
eholk: Is the unsafe block needed here? #Closed

rust-symcrypt/src/block_ciphers.rs
```diff
@@ -0,0 +1,17 @@
+use symcrypt_sys;
+
+
+// WIP still working on this
+pub enum BlockCipherType {
+    AES(symcrypt_sys::PCSYMCRYPT_BLOCKCIPHER),
```
eholk: This still ends up leaking the `PCSYMCRYPT_BLOCKCIPHER` type in the public API. I think what I'd do instead is just have just `AES`, and then have some private function that converts a `BlockCipherType` into a `PSYMCRYPT_BLOCKCIPHER`.

rust-symcrypt/src/chacha.rs
nnmkhang: the tag parameter in this case is an additional parameter that is outputed along with the encrypted result. The tag is used so that others can "check your work" and make sure that the encryption isnt tampered with in anyway. I can change to using the Tuple as that's what I had before.

rust-symcrypt/src/errors.rs
nnmkhang: Will add `#[non_exhaustive]`

rust-symcrypt/src/block_ciphers.rs
nnmkhang: Will change

rust-symcrypt/src/chacha.rs
nnmkhang: Changed return to `Result<(Vec<u8>, [u8; 16]), SymCryptError>`

rust-symcrypt/src/gcm.rs
nnmkhang: Going to go with returning a tuple, I think it is more clear and the trade off of memory allocation is just a sacrifice we'll make.

rust-symcrypt/src/gcm.rs
nnmkhang: done

rust-symcrypt/src/gcm.rs
nnmkhang: Not needed, removed

rust-symcrypt/src/block_ciphers.rs
nnmkhang: Changed

rust-symcrypt/src/errors.rs
```diff
@@ -0,0 +1,78 @@
...
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => SymCryptError::NoError,
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_UNUSED => SymCryptError::Unused,
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_WRONG_KEY_SIZE => SymCryptError::WrongKeySize,
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_WRONG_BLOCK_SIZE => SymCryptError::WrongBlockSize,
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_WRONG_DATA_SIZE => SymCryptError::WrongDataSize,
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_WRONG_NONCE_SIZE => SymCryptError::WrongNonceSize,
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_WRONG_TAG_SIZE => SymCryptError::WrongTagSize,
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_WRONG_ITERATION_COUNT => {
+                SymCryptError::WrongIterationCount
+            }
```
PhilHallin, line 45: Why the change in format?

nnmkhang: this is from cargo fmt

rust-symcrypt/src/chacha.rs
```diff
@@ -0,0 +1,115 @@
...
+            tag.len() as symcrypt_sys::SIZE_T,
+        ) {
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(dst),
+            err => Err(err.into()),
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
PhilHallin, line 78: Maybe add some tests that should return an error.

nnmkhang: will add

rust-symcrypt/src/chacha.rs
```diff
@@ -0,0 +1,132 @@
+/* ChaChaPoly1305 Functions. For further documentation please refer to symcrypt.h */
+
+use crate::errors::SymCryptError;
+use std::vec;
+use symcrypt_sys;
+
+pub fn chacha20_poly1305_encrypt(
+    key: &[u8],
```
mlindgren: For ChaCha20 Poly1305, the required key and nonce sizes are specified by the RFC:

- Key: 32 bytes
- Nonce: 12 bytes

I assume Rust will let you specify the size of these parameters, making it impossible to call this function with invalid arguments?

Edit: looks like you do this below for the tag input in decrypt. I'd suggest doing it for the key and nonce arguments as well.

rust-symcrypt/src/chacha.rs
```diff
@@ -0,0 +1,132 @@
...
+    nonce: &[u8],
+    auth_data: Option<&[u8]>,
+    src: &[u8],
```
mlindgren: I would call this `plaintext`. SymCrypt calls the buffers `src`/`dest` because your output buffer might be the same as your input buffer, but you're not supporting that here anyway. (Probably wouldn't be very idiomatic to do so in Rust.)

rust-symcrypt/src/chacha.rs
```diff
@@ -0,0 +1,132 @@
...
+            tag.as_mut_ptr(),
+            tag.len() as symcrypt_sys::SIZE_T,
+        ) {
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok((dst, tag)),
+            err => Err(err.into()),
+        }
+    }
+}
+
+pub fn chacha20_poly1305_decrypt(
```
mlindgren, line 42: Same comments as for encrypt

rust-symcrypt/src/gcm.rs
mlindgren: It is not sufficient to recalculate the magic values. They're designed to prevent callers from shooting themselves in the foot. From symcrypt.h:

```
// To provide quick error detection we have magic values in all
// our data structures, but only in CHKed builds.
// Our magic value depends on the address of the structure.
// This has the advantage that we detect blind memcpy's of our data structures.
// Memcpy is not supported as it limits what the library is allowed to do.
// Where needed the library provides for copy functions of its internal data structures.
```

We can't allow our structures to be copied because many of them have pointers to buffers, and it's not valid to copy those pointers without doing a deep copy of the underlying buffer. For example, if struct A has a pointer to buffer A, and gets memcpy'd to struct B, now struct B still points to buffer A. If struct A then gets freed, struct B is pointing to invalid memory, which can lead to a use-after-free. And there are many other things that could go wrong, e.g. some of our structs are self-referential.

Does Rust allow you to implement some kind of "copy" interface (e.g. copy constructor or `operator=` in C++) to ensure that the underlying data structures are copied appropriately? I think that would be the ideal way to solve this. Note, however, that some SymCrypt structures don't have associated copy functions.

Another way to solve this would be to not have the Rust structures directly contain the SymCrypt structures from bindgen, but rather to manually malloc space for the SymCrypt structures and have the rust structures only contain pointers to the SymCrypt structures. I guess that's kind of what Box does?

Pinning would probably also work.

This might need further discussion between the three of us.

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,409 @@
...
+    state: symcrypt_sys::SYMCRYPT_GCM_STATE,
+}
+
+// Have to Box<> since memory address for Self is moved around when returning from GcmState::new()
+// My guess is that it has to do with Result<> wrapping the Self and SymCryptError.
+// The only accepted Cipher for GCM is AesBlock.
+impl GcmState {
+    pub fn new(
+        key: &[u8],
+        nonce: &[u8],
```
mlindgren: Nonce size should be fixed to 12 bytes. This is particularly important here because the GCM functions don't do parameter validation except in debug builds, for performance reasons. From symcrypt.h:

```
// To achieve maximum performance, GCM functions do not check for valid parameters.
// Passing invalid parameters can lead to buffer overflows.
``` #Closed

File: rust-symcrypt/src/gcm.rs [hunk 20]
```diff
@@ -0,0 +1,409 @@
...
+        ) {
+            symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(dst),
+            err => Err(err.into()),
+        }
+    }
+}
+
+// SymCrypt requires pointer address to stay static through the scope, passing expand_key as an out
+// parameter to maintain static pointer address
+fn expand_key(
```
mlindgren: Should this be `gcm_expand_key`? Or does Rust have namespaces so that all of these standalone functions could be a GCM namespace? #Closed

symcrypt-sys/build.rs
```diff
@@ -48,13 +48,15 @@ fn main() {
         // HMAC FUNCTIONS
         .allowlist_function("^(SymCryptHmacSha256.*)$")
         .allowlist_function("^(SymCryptHmacSha384.*)$")
-        // GCM FUNCTIONS
+        // CM FUNCTIONS
```
mlindgren: ?

rust-symcrypt/src/gcm.rs
mlindgren: Likewise for other functions that take nonce values

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,409 @@
...
+                &mut instance.state,
+                &instance.expanded_key,
+                nonce.as_ptr(),
+                nonce.len() as symcrypt_sys::SIZE_T,
+            );
+            Ok(instance)
+        }
+    }
+
+    pub fn auth(&mut self, auth_data: &[u8]) {
```
mlindgren:
```suggestion
    pub fn auth_part(&mut self, auth_data: &[u8]) {
```
to be consistent with others, and to indicate that this function can be called repeatedly #Closed

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,409 @@
...
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptGcmAuthPart(
+                &mut self.state,
+                auth_data.as_ptr(),
+                auth_data.len() as symcrypt_sys::SIZE_T,
+            );
+        }
+    }
+
+    pub fn encrypt_part(&mut self, data: &[u8]) -> Vec<u8> {
```
mlindgren: This will potentially incur a lot of copying when encrypting large amounts of data, as you need to put all of the pieces together at the end. I know we've gone back and forth on how much to write the code for performance versus idiomatic Rust design, but I think the performance impact here could be a lot higher than in the other cases we've discussed because the copying will probably take vastly more time than the encryption itself.

Is there a way to take a slice of an existing array/vector as a mutable input and write the result to that?

If you want to leave it as-is for now and make a decision based on benchmarking it later, I'm fine with that. But I do expect that there will be significant performance impact.

Same comment for decrypt. #Closed

mlindgren (changes requested):
Left a few suggestions. In particular, we should discuss more about how to handle structures that can't be copied or moved.

rust-symcrypt/src/gcm.rs
PhilHallin: Maybe add another function, such as fast_encrypt_part() that updates an output parameter instead of returning a result..

rust-symcrypt/src/chacha.rs
nnmkhang: Will make the update for nonce.

rust-symcrypt/src/gcm.rs
nnmkhang: This is a private function and wont be accessible to other modules but I will change to `gcm_expand_key `for clarity.

symcrypt-sys/build.rs
nnmkhang: Thought I needed to support other CM functions, will change back to GCM.

rust-symcrypt/src/gcm.rs
PhilHallin: Yes, Box is ensuring that only one instance of the object is allocated on the heap.

rust-symcrypt/src/gcm.rs
nnmkhang: Will change

rust-symcrypt/src/gcm.rs
nnmkhang: We can take in an existing slice and modify it, similar to how `expand_key()` works. I am open to changing it over now, will track this as a discussion item

PhilHallin: Might also want to do for the chacha encrypt/decrypt functions.


The pattern for encrypt/decrypt would be to do as an out parameter.

nnmkhang: > Maybe add another function, such as fast_encrypt_part() that updates an output parameter instead of returning a result..

I dont see a point in doing a fast version, Inherently this is already supposed to be fast, if the speed is an issue I think we have to just move over to the out parameter.

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,455 @@
...
+    // Can be called multiple times
+    // Takes in a buffer as an out parameter for performance reasons
+    pub fn decrypt_part(&mut self, cipher_text: &[u8], plain_text_buffer: &mut [u8]) {
+        unsafe {
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptGcmDecryptPart(
+                &mut self.inner.state,
+                cipher_text.as_ptr(),
+                plain_text_buffer.as_mut_ptr(),
+                plain_text_buffer.len() as symcrypt_sys::SIZE_T,
```
PhilHallin, line 119: >plain_text_buffer.len() as symcrypt_sys::SIZE_T, [](http://example.com/codeflow?start=16&length=48)

Need to check that cipher_text.len() == plain_text_buffer.len() #Closed

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,455 @@
...
+struct GcmStateInner {
+    state: symcrypt_sys::SYMCRYPT_GCM_STATE,
+    expanded_key: symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY,
+}
+
+/// SymCrypt expects the address for its structs to stay static through the structs lifetime.
+/// Using an inner GCM state that is Pin<Box<T>> since the memory address for Self is moved around when returning from GcmState::new()
+/// My guess is that it has to do with Result<> wrapping the Self and SymCryptError.
+///
+/// auth_part, encrypt_part, decrypt_part take in an allocated buffer as an out parameter for performance reasons. This is for scenarios
```
PhilHallin: >auth_part [](http://example.com/codeflow?start=4&length=9)

auth_part only takes as an input parameter.

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,455 @@
...
+    // Can be called multiple times
+    // Takes in a buffer as an out parameter for performance reasons
+    pub fn encrypt_part(&mut self, plain_text: &[u8], cipher_text_buffer: &mut [u8]) {
+        unsafe {
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptGcmEncryptPart(
+                &mut self.inner.state,
+                plain_text.as_ptr(),
+                cipher_text_buffer.as_mut_ptr(),
+                cipher_text_buffer.len() as symcrypt_sys::SIZE_T,
```
PhilHallin, line 89: >cipher_text_buffer.len() as symcrypt_sys::SIZE_T, [](http://example.com/codeflow?start=16&length=49)

Need to check that plain_text.len() == cipher_text_buffer.len() #Closed

PhilHallin: Instead of returning an error, might be OK to panic instead. It is a caller mistake if they aren't ==

Question for Eric?

rust-symcrypt/src/gcm.rs
```diff
@@ -0,0 +1,477 @@
...
+    #[test]
+    fn test_stateless_gcm_encrypt_ad() {
+        let p_key = hex::decode("feffe9928665731c6d6a8f9467308308").unwrap();
+        let nonce = hex::decode("cafebabefacedbaddecaf888").unwrap();
+        let auth_data = hex::decode("feedfacedeadbeeffeedfacedeadbeefabaddad2").unwrap();
+        let expected_tag = "5bc94fbc3221a5db94fae95ae7121a47";
+        let expected_result = "42831ec2217774244b7221b784d0d49ce3aa212f2c02a4e035c17e2329aca12e21d514b25466931c7d8f6a5aac84aa051ba30b396a0aac973d58e091";
+        let pt = hex::decode("d9313225f88406e5a55909c5aff5269a86a7a9531534f7da2e4c303d8a318a721c3c0c95956809532fcf0e2449a6b525b16aedf5aa0de657ba637b39").unwrap();
+
+        let nonce_array: [u8; 12] = nonce.as_slice().try_into().unwrap();
```
PhilHallin: >let nonce_array: [u8; 12] = nonce.as_slice().try_into().unwrap(); [](http://example.com/codeflow?start=8&length=65)

Why is this necessary?  Can't you do the following:

&nonce[0 .. 12] #Closed

nnmkhang: `&nonce[0 .. 12]` produces `&[u8]`, we need ` [u8; 12] `

rust-symcrypt/src/chacha.rs
```diff
@@ -88,7 +88,11 @@ mod test {
         let ct = "d31a8d34648e60db7b86afbc53ef7ec2a4aded51296e08fea9e2b5a736ee62d63dbea45e8ca9671282fafb69da92728b1a71de0a9e060b2905d6a5b67ecd3b3692ddbd7f2d778b8c9803aee328091b58fab324e4fad675945585808b4831d7bc3ff4def08e4b7a9de576d26586cec64b6116";
         let expected_tag = "1ae10b594f09e26a7e902ecbd0600691";
 
-        let (result, tag) = chacha20_poly1305_encrypt(&key, &nonce, Some(&auth_data), &pt).unwrap();
+        let nonce_array: [u8; 12] = nonce.as_slice().try_into().unwrap();
```
mlindgren: Probably me not understanding Rust... why can't this be passed directly?

rust-symcrypt/src/gcm.rs
```diff
@@ -4,97 +4,119 @@ use crate::block_ciphers::*;
...
 use core::ffi::c_void;
 use std::mem;
+use std::pin::Pin;
 use std::ptr;
 use std::vec;
 use symcrypt_sys;
 
 pub struct GcmState {
-    expanded_key: symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY,
+    inner: Pin<Box<GcmStateInner>>,
```
mlindgren, line 13: Doesn't need to be changed for this PR, but I think you'll have to go back and update the other Rust structures you've created to use this same pattern (e.g. for hash, HMAC). Those SymCrypt structures are also designed to not be `memcpy`'d, and also have magic fields. I'm not sure why you didn't hit errors before with those. #Closed

rust-symcrypt/src/gcm.rs
```diff
@@ -4,97 +4,119 @@ use crate::block_ciphers::*;
...
+
+struct GcmStateInner {
     state: symcrypt_sys::SYMCRYPT_GCM_STATE,
+    expanded_key: symcrypt_sys::SYMCRYPT_GCM_EXPANDED_KEY,
 }
 
-// Have to Box<> since memory address for Self is moved around when returning from GcmState::new()
-// My guess is that it has to do with Result<> wrapping the Self and SymCryptError.
-// The only accepted Cipher for GCM is AesBlock.
+/// SymCrypt expects the address for its structs to stay static through the structs lifetime.
```
mlindgren: Possibly worth noting in this comment that the reason for this is that to guarantee that the structs aren't `memcpy`'d as doing so can lead to use-after-free, inconsistent states, etc. #Closed

rust-symcrypt/src/gcm.rs
```diff
@@ -4,97 +4,119 @@ use crate::block_ciphers::*;
...
+/// Using an inner GCM state that is Pin<Box<T>> since the memory address for Self is moved around when returning from GcmState::new()
+/// My guess is that it has to do with Result<> wrapping the Self and SymCryptError.
```
mlindgren: I'm not sure if it's due to the `Result` type, but rather just because `instance` is allocated on the stack, so when the function returns, the values have to be copied to their destination. In some cases the compiler might be smart enough to optimize this out, but it probably can't do so all the time.

(That said, standard caveat that I don't know much about Rust.) #Closed

mlindgren (approved):

rust-symcrypt/src/gcm.rs
PhilHallin: // Create a mutable array of 12 bytes
let mut nonce: [u8; 12] = [0; 12];
// Decode the hex string and write the result into the array
hex::decode_to_slice("cafebabefacedbaddecaf888", &mut nonce).unwrap();

PhilHallin: hex::decode creates the Vec

rust-symcrypt/src/gcm.rs
nnmkhang: Have it tracked as an issue

rust-symcrypt/src/gcm.rs
nnmkhang: Will make note

rust-symcrypt/src/chacha.rs
nnmkhang: `nonce` is a `Vec<u8> ` type, which is different from an array slice that forced to be 12 bytes. `&[u8; 12]`.  ( `Vec<T>` is a heap allocated and the memory is dynamically created ). Thats why we have to do `try_into()` to see if it can fit into 12 bytes, because if it wont the compiler will `panic() `and crash.

rust-symcrypt/src/gcm.rs
nnmkhang: I had the same idea as well, when I tired this on other code like HMAC, there was no issue with the magic. The magic issue only started occurring once I introduced the Result type. Its entirely possible that we were just getting lucky before and the compiler was able to do some in line move to make the memory address match up.

I will remove my speculation for clarity

PhilHallin (approved):
:shipit:

PR #20, opened 2023-08-15
Adding EcDh and needed support code

nnmkhang:
Added:
- ecdh.rs
- curve_types.rs

Modified:
- build.rs
- symcrypt_bindings.rs

### DIFF
--- a/rust-symcrypt/src/ecdh.rs
+++ b/rust-symcrypt/src/ecdh.rs
@@ -6,15 +6,15 @@ use std::vec;
 use symcrypt_sys;
 
 /// EcDhKey is a wrapper around symcrypt_sys::PSYMCRYPT_ECKEY, this is to let rust handle the free'ing of this pointer
-/// EcDhKey must be allocated after EcDhExpandedCurve and free'd before EcDhExpandedCurve is free'd
+/// EcDhKey must be allocated after EcDhCurve and free'd before EcDhCurve is free'd
 struct EcDhKey {
     inner: symcrypt_sys::PSYMCRYPT_ECKEY,
-    expanded_curve: EcDhExpandedCurve,
+    curve: EcDhCurve,
 }
 
-struct EcDhExpandedCurve(symcrypt_sys::PSYMCRYPT_ECURVE);
+struct EcDhCurve(symcrypt_sys::PSYMCRYPT_ECURVE);
 
-impl Drop for EcDhExpandedCurve {
+impl Drop for EcDhCurve {
     fn drop(&mut self) {
         unsafe { symcrypt_sys::SymCryptEcurveFree(self.0) }
     }
@@ -23,13 +23,13 @@ impl Drop for EcDhExpandedCurve {
 impl EcDhKey {
     pub fn new(curve: CurveType) -> Result<Self, SymCryptError> {
         unsafe {
-            // Allocating expanded_curve needed for key derivation. This pointer must be dropped after the key is dropped.
+            // Allocating curve needed for key derivation. This pointer must be dropped after the key is dropped.
             let curve_ptr = symcrypt_sys::SymCryptEcurveAllocate(convert_curve(curve), 0);
             if curve_ptr.is_null() {
                 return Err(SymCryptError::MemoryAllocationFailure);
             }
-            // expanded_curve needs to be wrapped to properly free the curve in the case there is an error in EcDhKey initialization
-            let expanded_curve = EcDhExpandedCurve(curve_ptr);
+            // curve needs to be wrapped to properly free the curve in the case there is an error in EcDhKey initialization
+            let ecdh_curve = EcDhCurve(curve_ptr);
 
             // Key must be dropped before curve is dropped
             let key_ptr = symcrypt_sys::SymCryptEckeyAllocate(curve_ptr);
@@ -38,7 +38,7 @@ impl EcDhKey {
             }
             let key = EcDhKey {
                 inner: key_ptr,
-                expanded_curve: expanded_curve,
+                curve: ecdh_curve,
             };
 
             Ok(key)
@@ -150,16 +150,13 @@ impl EcDh {
     }
 
     pub fn ecdh_secret_agreement(private: &EcDh, public: &EcDh) -> Result<Vec<u8>, SymCryptError> {
-        if private.curve_type != public.curve_type {
-            return Err(SymCryptError::InvalidArgument);
-        }
-
+        
         let num_format = get_num_format(private.curve_type);
 
         unsafe {
             // SAFETY: FFI calls
             let secret_length =
-                symcrypt_sys::SymCryptEcurveSizeofFieldElement(private.key.expanded_curve.0);
+                symcrypt_sys::SymCryptEcurveSizeofFieldElement(private.key.curve.0);
             let mut secret = vec![0u8; secret_length as usize];
 
             match symcrypt_sys::SymCryptEcDhSecretAgreement(
--- a/rust-symcrypt/src/ecdh.rs
+++ b/rust-symcrypt/src/ecdh.rs
@@ -37,15 +37,19 @@ impl EcDh {
 
             instance.key = Some(key_ptr);
 
-            symcrypt_sys::SymCryptEckeySetRandom(
+            match symcrypt_sys::SymCryptEckeySetRandom(
                 symcrypt_sys::SYMCRYPT_FLAG_ECKEY_ECDH,
                 instance.key.unwrap(),
-            );
+            ) {
+                symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => {
+                    Ok(instance)
+                }
+                err => Err(err.into()),
+            }
         }
-        Ok(instance)
     }
 
-    /// ExDh::from_public_key_bytes function returns a EcDh struct that only has a public key attached.
+    /// EcDh::from_public_key_bytes()    function returns a EcDh struct that only has a public key attached.
     pub fn from_public_key_bytes(
         curve: CurveType,
         public_key: &[u8],
@@ -56,11 +60,7 @@ impl EcDh {
             key: None,
         };
 
-        let num_format = if instance.curve_type == CurveType::Curve25519 {
-            symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_LSB_FIRST
-        } else {
-            symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_MSB_FIRST
-        };
+        let num_format = get_num_format(instance.curve_type);
 
         let ec_point_format = symcrypt_sys::_SYMCRYPT_ECPOINT_FORMAT_SYMCRYPT_ECPOINT_FORMAT_XY;
 
@@ -76,11 +76,11 @@ impl EcDh {
 
             let key_ptr = symcrypt_sys::SymCryptEckeyAllocate(instance.expanded_curve.unwrap());
             if key_ptr.is_null() {
-                return Err(SymCryptError::AuthenticationFailure);
+                return Err(SymCryptError::InvalidArgument);
             }
 
             match symcrypt_sys::SymCryptEckeySetValue(
-                std::ptr::null_mut(),
+                std::ptr::null(),
                 0,
                 public_key.as_ptr(),
                 public_key.len() as symcrypt_sys::SIZE_T,
@@ -99,12 +99,7 @@ impl EcDh {
     }
 
     pub fn get_public_key_bytes(&mut self) -> Result<Vec<u8>, SymCryptError> {
-        let num_format = if self.curve_type == CurveType::Curve25519 {
-            symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_LSB_FIRST
-        } else {
-            symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_MSB_FIRST
-        };
-
+        let num_format = get_num_format(self.curve_type);
         let ec_point_format = symcrypt_sys::_SYMCRYPT_ECPOINT_FORMAT_SYMCRYPT_ECPOINT_FORMAT_XY;
 
         unsafe {
@@ -134,14 +129,10 @@ impl EcDh {
 
     pub fn ecdh_secret_agreement(private: &EcDh, public: &EcDh) -> Result<Vec<u8>, SymCryptError> {
         if private.curve_type != public.curve_type {
-            return Err(SymCryptError::AuthenticationFailure);
+            return Err(SymCryptError::InvalidArgument);
         }
 
-        let num_format = if private.curve_type == CurveType::Curve25519 {
-            symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_LSB_FIRST
-        } else {
-            symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_MSB_FIRST
-        };
+        let num_format = get_num_format(private.curve_type);
 
         unsafe {
             // SAFETY: FFI calls
@@ -178,19 +169,23 @@ impl Drop for EcDh {
     }
 }
 
+fn get_num_format(curve_type: CurveType)-> i32 {
+    if curve_type == CurveType::Curve25519 {
+        return symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_LSB_FIRST;
+    } else {
+        return symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_MSB_FIRST;
+    };
+}
+
 mod test {
     use super::*;
-    use crate::init;
+    use crate::symcrypt_init;
 
     /// symcrypt_sys::SymCryptModuleInit() must be called via lib.rs in order to initialize the callbacks for
     /// SymCryptEcurveAllocate, SymCryptEckeyAllocate, SymCryptCallbackAlloc, etc.
-    /// Calling init() at the beginning of each test function is causing a mismatch error in the magic, so we
-    /// are wrapping all tests under one function to call the init() only once. This is closer to the actual
-    /// behavior of symcrypt since SymCryptModuleInit() would only be called once at the beginning of the module.
-
     #[test]
     fn test_ecdh_nist_p256() {
-        init(); // must run init for the alloc callbacks to initialize
+        symcrypt_init(); // must run symcrypt_init for the alloc callbacks to initialize
         let mut ecdh_1_private = EcDh::new(CurveType::NistP256).unwrap();
         let mut ecdh_2_private = EcDh::new(CurveType::NistP256).unwrap();
 
@@ -212,7 +207,7 @@ mod test {
 
     #[test]
     fn test_ecdh_nist_p384() {
-        init(); // must run init for the alloc callbacks to initialize
+        symcrypt_init(); // must run symcrypt_init for the alloc callbacks to initialize
         let mut ecdh_1_private = EcDh::new(CurveType::NistP384).unwrap();
         let mut ecdh_2_private = EcDh::new(CurveType::NistP384).unwrap();
 
@@ -234,7 +229,7 @@ mod test {
 
     #[test]
     fn test_ecdh_curve_25519() {
-        init(); // must run init for the alloc callbacks to initialize
+        symcrypt_init(); // must run symcrypt_init for the alloc callbacks to initialize
         let mut ecdh_1_private = EcDh::new(CurveType::Curve25519).unwrap();
         let mut ecdh_2_private = EcDh::new(CurveType::Curve25519).unwrap();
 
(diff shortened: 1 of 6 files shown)


PhilHallin:
> }

This should be executed once via using

std::Sync::Once #Closed

---
Refers to: rust-symcrypt/src/lib.rs:8 in d214a7f. [](commit_id = d214a7f9f7a0d39dc853d766241b98bb59e55e4a, deletion_comment = False)

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,352 @@
...
+
+mod test {
+    use super::*;
+    use crate::init;
+
+    /// symcrypt_sys::SymCryptModuleInit() must be called via lib.rs in order to initialize the callbacks for
+    /// SymCryptEcurveAllocate, SymCryptEckeyAllocate, SymCryptCallbackAlloc, etc.
+    /// Calling init() at the beginning of each test function is causing a mismatch error in the magic, so we
+    /// are wrapping all tests under one function to call the init() only once. This is closer to the actual
+    /// behavior of symcrypt since SymCryptModuleInit() would only be called once at the beginning of the module.
```
PhilHallin: Updating your init() to use std::Sync::Once() should address this.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,352 @@
...
+            instance.expanded_curve = Some(curve_ptr);
+
+            let key_ptr = symcrypt_sys::SymCryptEckeyAllocate(curve_ptr);
+            if key_ptr.is_null() {
+                return Err(SymCryptError::MemoryAllocationFailure);
+            }
+
+            instance.key = Some(key_ptr);
+
+            symcrypt_sys::SymCryptEckeySetRandom(
```
PhilHallin: >symcrypt_sys::SymCryptEckeySetRandom [](http://example.com/codeflow?start=12&length=36)

This API can fail with SYMCRYPT_ERROR which needs to be returned. Can't panic for an error. #Closed

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,352 @@
...
+                symcrypt_sys::SYMCRYPT_FLAG_ECKEY_ECDH,
+                instance.key.unwrap(),
```
PhilHallin: > instance.key.unwrap() [](http://example.com/codeflow?start=15&length=22)

Why the unwrap()? Could you just use key_ptr?

rust-symcrypt/src/ecdh.rs
PhilHallin: Do same as you are doing in from_public_key_bytes

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,352 @@
...
+
+            let key_ptr = symcrypt_sys::SymCryptEckeyAllocate(instance.expanded_curve.unwrap());
+            if key_ptr.is_null() {
+                return Err(SymCryptError::AuthenticationFailure);
+            }
+
+            match symcrypt_sys::SymCryptEckeySetValue(
+                std::ptr::null_mut(),
+                0,
+                public_key.as_ptr(),
```
PhilHallin, line 96: >public_key.as_ptr() [](http://example.com/codeflow?start=16&length=19)

Just curious. What is as_ptr() for an empty slice?

Shouldn't be a problem, where .len() will be 0 so it shouldn't be used?

rust-symcrypt/src/ecdh.rs
PhilHallin: I see. Use .unwrap for the Option()

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,352 @@
...
+
+            instance.expanded_curve = Some(curve_ptr);
+
+            let key_ptr = symcrypt_sys::SymCryptEckeyAllocate(instance.expanded_curve.unwrap());
+            if key_ptr.is_null() {
+                return Err(SymCryptError::AuthenticationFailure);
+            }
+
+            match symcrypt_sys::SymCryptEckeySetValue(
+                std::ptr::null_mut(),
```
PhilHallin: >mut [](http://example.com/codeflow?start=31&length=3)

Does this need to be _mut? Its PCBYTE #Closed

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,352 @@
...
+            curve_type: curve,
+            expanded_curve: None,
+            key: None,
+        };
+
+        let num_format = if instance.curve_type == CurveType::Curve25519 {
+            symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_LSB_FIRST
+        } else {
+            symcrypt_sys::_SYMCRYPT_NUMBER_FORMAT_SYMCRYPT_NUMBER_FORMAT_MSB_FIRST
+        };
```
PhilHallin: Maybe a shared function? #Closed

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,352 @@
...
+            ) {
+                symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(pub_key_bytes),
+                err => Err(err.into()),
+            }
+        }
+    }
+
+    pub fn ecdh_secret_agreement(private: &EcDh, public: &EcDh) -> Result<Vec<u8>, SymCryptError> {
+        if private.curve_type != public.curve_type {
+            return Err(SymCryptError::AuthenticationFailure);
```
PhilHallin: >AuthenticationFailure [](http://example.com/codeflow?start=38&length=21)

Maybe InvalidArgument #Closed

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,352 @@
...
+    fn test_ecdh_nist_p256() {
+        init(); // must run init for the alloc callbacks to initialize
+        let mut ecdh_1_private = EcDh::new(CurveType::NistP256).unwrap();
+        let mut ecdh_2_private = EcDh::new(CurveType::NistP256).unwrap();
+
+        let public_bytes_1 = ecdh_1_private.get_public_key_bytes().unwrap();
+        let public_bytes_2 = ecdh_2_private.get_public_key_bytes().unwrap();
+
+        let ecdh_1_public =
+            EcDh::from_public_key_bytes(CurveType::NistP256, &public_bytes_1.as_slice()).unwrap();
```
PhilHallin, line 187: >as_slice() [](http://example.com/codeflow?start=77&length=10)

Is this necessary?

Can't you just pass as &public_bytes_1

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,269 @@
...
+                symcrypt_sys::SymCryptEckeyFree(key);
+            }
+            if let Some(expanded_curve) = self.expanded_curve {
+                symcrypt_sys::SymCryptEcurveFree(expanded_curve);
+            }
+        }
+    }
+}
+
+fn get_num_format(curve_type: CurveType)-> i32 {
```
PhilHallin: Seems like this should be in curve_type.rs?

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,269 @@
...
+
+        unsafe {
+            // SAFETY: FFI calls
+            let curve_ptr =
+                symcrypt_sys::SymCryptEcurveAllocate(convert_curve(instance.curve_type), 0);
+            if curve_ptr.is_null() {
+                return Err(SymCryptError::MemoryAllocationFailure);
+            }
+
+            instance.expanded_curve = Some(curve_ptr);
```
eholk: Can `expanded_curve` every be `None` from this point on? If not, I'd recommend structuring this function sort of like:

```rust 
let curve_type = curve;
let expanded_curve = SymCryptEcurveAllocate(...);
let mut key_ptr = ptr::null();
let key = match SymCryptEckeySetValue(...) {
    NO_ERROR => key_ptr,
    err => return Err(err.into()),
}

EcDh {
    curve_type,
    expanded_curve,
    key
}
```

This way you don't have to worry about unwrapping the fields later on since they are guaranteed always to be valid.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,269 @@
...
+                ec_point_format,
+                0, // no flags allowed
+            ) {
+                symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(pub_key_bytes),
+                err => Err(err.into()),
+            }
+        }
+    }
+
+    pub fn ecdh_secret_agreement(private: &EcDh, public: &EcDh) -> Result<Vec<u8>, SymCryptError> {
```
eholk: Instead of returning `Vec<u8>` here, I'd consider returning something like `SecreteAgreement` which would be a type that wraps the array of bytes. It makes it easier to keep track of what the meaning of the bytes is elsewhere in the code.

rust-symcrypt/src/ecdh.rs
PhilHallin: In the above err => we would also need to call 
symcrypt_sys::SymCryptEcurveFree(expanded_curve)

PhilHallin: Also, we were using the instance data structure to hold the pointers that would get freed for any errors in the SymCryptEckeySetValue call.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,269 @@
...
+            if curve_ptr.is_null() {
+                return Err(SymCryptError::MemoryAllocationFailure);
+            }
+
+            instance.expanded_curve = Some(curve_ptr);
+
+            let key_ptr = symcrypt_sys::SymCryptEckeyAllocate(instance.expanded_curve.unwrap());
+            if key_ptr.is_null() {
+                return Err(SymCryptError::InvalidArgument);
+            }
```
PhilHallin: >} [](http://example.com/codeflow?start=12&length=1)

instance.key needs to be updated here so the key gets freed for any subsequent errors.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,270 @@
...
+use crate::curve_type::*;
+use crate::errors::SymCryptError;
+use std::vec;
+use symcrypt_sys;
+
+/// EcDhKey is a wrapper around symcrypt_sys::PSYMCRYPT_ECKEY, this is to let rust handle the free'ing of this pointer
+/// EcDhKey must be allocated after EcDhExpandedCurve and free'd before EcDhExpandedCurve is free'd
+struct EcDhKey {
+    inner: symcrypt_sys::PSYMCRYPT_ECKEY,
+    expanded_curve: EcDhExpandedCurve,
```
eholk: Is it ever possible for multiple keys to share the same expanded curve?

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,270 @@
...
+impl EcDhKey {
+    pub fn new(curve: CurveType) -> Result<Self, SymCryptError> {
+        unsafe {
+            // Allocating expanded_curve needed for key derivation. This pointer must be dropped after the key is dropped.
+            let curve_ptr = symcrypt_sys::SymCryptEcurveAllocate(convert_curve(curve), 0);
+            if curve_ptr.is_null() {
+                return Err(SymCryptError::MemoryAllocationFailure);
+            }
+            // expanded_curve needs to be wrapped to properly free the curve in the case there is an error in EcDhKey initialization
+            let expanded_curve = EcDhExpandedCurve(curve_ptr);
```
eholk: I'd factor the `SymCryptEcurveAllocate` code into `EcDhExpandedCurve::new` and then replace the last line here with `let expanded_curve = EcDhExpandedCurve::new()?;` but it doesn't really matter since these are all internal implementation details.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,270 @@
...
+
+use crate::curve_type::*;
+use crate::errors::SymCryptError;
+use std::vec;
+use symcrypt_sys;
+
+/// EcDhKey is a wrapper around symcrypt_sys::PSYMCRYPT_ECKEY, this is to let rust handle the free'ing of this pointer
+/// EcDhKey must be allocated after EcDhExpandedCurve and free'd before EcDhExpandedCurve is free'd
+struct EcDhKey {
+    inner: symcrypt_sys::PSYMCRYPT_ECKEY,
```
eholk, line 11:
```suggestion
    inner: NonNull<symcrypt_sys::SYMCRYPT_ECKEY>,
```

You guarantee by construction that this pointer is never null, so you can put that in the type system if you want. It doesn't make a huge difference since you don't have any null checks on this anyway though.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,270 @@
...
+    expanded_curve: EcDhExpandedCurve,
+}
+
+struct EcDhExpandedCurve(symcrypt_sys::PSYMCRYPT_ECURVE);
```
eholk:
```suggestion
struct EcDhExpandedCurve(NonNull<symcrypt_sys::SYMCRYPT_ECURVE>);
```

Same thing here, you only ever construct this with non null pointers, so you can tell the type system that's the case.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,270 @@
...
+            let expanded_curve = EcDhExpandedCurve(curve_ptr);
+
+            // Key must be dropped before curve is dropped
+            let key_ptr = symcrypt_sys::SymCryptEckeyAllocate(curve_ptr);
+            if key_ptr.is_null() {
+                return Err(SymCryptError::MemoryAllocationFailure);
+            }
+            let key = EcDhKey {
+                inner: key_ptr,
+                expanded_curve: expanded_curve,
```
eholk:
```suggestion
                expanded_curve,
```

This is more a style suggestion, so I'll leave it up to your team, but you can just say `expanded_curve` here and that's sugar for `expanded_curve: expanded_curve`.

eholk (commented):
From a Rust standpoint, this looks pretty good! I left a few suggestions, but none of them are blocking.

rust-symcrypt/src/ecdh.rs
mlindgren: Yes, it is. The key does not own the curve; it is the responsibility of whoever called `SymCryptEcurveAllocate` to clean it up by calling `SymCryptEcurveFree`

Edit: it's possible via the SymCrypt APIs, at least. For the Rust wrappers, we could guarantee that there's a one-to-one mapping between key objects and curve objects. This could result in a lot of extra memory use if you're using many different EC keys, though.

rust-symcrypt/src/ecdh.rs
mlindgren: >EcDhExpandedCurve [](http://example.com/codeflow?start=7&length=17)

Why "expanded"? Also, `PSYMCRYPT_ECURVE` is used for other elliptic curve algorithms, so I'm not sure it makes sense to wrap it with a structure that is named specifically for ECDH. #Resolved

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,270 @@
...
+            };
+
+            Ok(key)
+        }
+    }
+}
+
+/// Must drop the Key before the expanded curve is dropped.
```
mlindgren, line 44: Does the `drop` function for the "outer" structure always run before the `drop` function for the "inner" structure? #Resolved

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,270 @@
...
+            ) {
+                symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => Ok(pub_key_bytes),
+                err => Err(err.into()),
+            }
+        }
+    }
+
+    pub fn ecdh_secret_agreement(private: &EcDh, public: &EcDh) -> Result<Vec<u8>, SymCryptError> {
+        if private.curve_type != public.curve_type {
+            return Err(SymCryptError::InvalidArgument);
```
mlindgren: >return Err(SymCryptError [](http://example.com/codeflow?start=12&length=24)

SymCrypt will validate that the curves are the same for you. I think it's probably best to just rely on SymCrypt for that and not duplicate parts of the logic here. #Resolved

rust-symcrypt/src/ecdh.rs
mlindgren: +1

rust-symcrypt/src/ecdh.rs
mlindgren: +1

rust-symcrypt/src/ecdh.rs
eholk: Does the curve object have a ref count internally? If so, we could reflect that here to do shared ownership of the keys.

If not, another option is to have the Rust curve wrapper use an `Arc` or `Rc` and add shared ownership on top of the underlying SymCrypt structures. That means we'd have an extra allocation and a layer of indirection though, so it may not be worthwhile.

rust-symcrypt/src/ecdh.rs
eholk: Yes, the outer destructors run first and then inner destructors run.

https://doc.rust-lang.org/reference/destructors.html

rust-symcrypt/src/ecdh.rs
mlindgren: No, it’s not refcounted by SymCrypt.

PhilHallin: Khang is planning on using lazy static to call SymCryptEcurveAllocate(curve) on first use and share for subsequent calls.

eholk: Would there be any problem with sharing one curve for the whole process? Are there use cases where it makes sense to have two different curves allocated?

PhilHallin: Should only need one instance per curve type per process.

I'm seeing in SymCrypt-OpenSSL it calls SymCryptEcurveAllocate() for each supported curve type during engine init.

We will be better in rust, where we defer until the curve is actually used.

mlindgren: Yeah, should be fine to just use one curve. I don't know why SymCrypt doesn't just hardcode the values for well-known curves. If I had to guess it's to save on binary size for applications that don't use ECC, especially since we want to support embedded devices. But I'd have to ask Niels to be sure.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,257 @@
...
+            EcDh::from_public_key_bytes(CurveType::NistP256, &public_bytes_1.as_slice()).unwrap();
+        let ecdh_2_public =
+            EcDh::from_public_key_bytes(CurveType::NistP256, &public_bytes_2.as_slice()).unwrap();
+
+        let secret_agreement_1 =
+            EcDh::ecdh_secret_agreement(&ecdh_1_private, &ecdh_2_public).unwrap();
+        let secret_agreement_2 =
+            EcDh::ecdh_secret_agreement(&ecdh_2_private, &ecdh_1_public).unwrap();
+
+        assert_eq!(secret_agreement_1.0, secret_agreement_2.0);
```
mlindgren: >assert_eq!(secret_agreement_1.0, secret_agreement_2.0); [](http://example.com/codeflow?start=8&length=55)

Can `EcDhSecretAgreement` objects not be compared directly? They're basically just wrappers/typedefs/aliases (not sure the correct Rust terminology) for vectors, so it seems like that should work...

rust-symcrypt/src/ecurve.rs
```diff
@@ -0,0 +1,76 @@
...
+
+pub(crate) fn convert_curve(curve: CurveType) -> symcrypt_sys::PCSYMCRYPT_ECURVE_PARAMS {
+    match curve {
+        CurveType::NistP256 => unsafe { symcrypt_sys::SymCryptEcurveParamsNistP256 }, // SAFETY: FFI calls
+        CurveType::NistP384 => unsafe { symcrypt_sys::SymCryptEcurveParamsNistP384 }, // SAFETY: FFI calls
+        CurveType::Curve25519 => unsafe { symcrypt_sys::SymCryptEcurveParamsCurve25519 }, // SAFETY: FFI calls
+    }
+}
+
+pub(crate) fn get_num_format(curve_type: CurveType) -> i32 {
```
mlindgren, line 23: >get_num_format [](http://example.com/codeflow?start=14&length=14)

(Mentioned this on Teams, but copying it here for posterity)

I was trying to understand why you're using a different format for Curve25519. I looked it up and I guess Curve25519 typically uses LE. But the `SYMCRYPT_NUMBER_FORMAT` options are only used for serializing and deserializing, and do not affect SymCrypt's internal data structures. So either option should be fine for any curve.

I think this is fine as-is, but maybe leave a comment explaining why you wrote it this way.

mlindgren (approved):
:shipit:

rust-symcrypt/src/ecurve.rs
PhilHallin: This is called for the following APIs:
SymCryptEckeySetValue, SymCryptEckeyGetValue and SymCryptEcDhSecretAgreement.

Each of which have the following input parameter:
SYMCRYPT_NUMBER_FORMAT  format,

mlindgren: Right, but the parameter refers to the endianness of the input/output buffers. It doesn’t directly correspond to which curve is being used.

PhilHallin: In the file: \onecore\ds\security\cryptoapi\ncrypt\msprim\ecc\ecc.c, I'm seeing the following for this parameter for the APIs listed above:

IsCurve25519(pKey->pParameters) ? SYMCRYPT_NUMBER_FORMAT_LSB_FIRST : SYMCRYPT_NUMBER_FORMAT_MSB_FIRST,

PhilHallin: We will be importing/exporting the 25519 public key using these APIs. What is the format on the wire, LSB or MSB?

mlindgren: Ah, interesting. I don't know why BCrypt does that either. Possibly for the same reason I mentioned above, that Curve25519 conventionally uses LSB first. But I don't know if the format BCrypt/SymCrypt formats are directly compatible with those used by other implementations, e.g. OpenSSL. I'll ask Niels about this (or you can if you run into him first.)

rust-symcrypt/src/ecdh.rs
eholk: You have to opt in to being able to compare them. A lot of types this kind of newtype wrapper is done specifically to remove the ability to compare by equality. But if you add `#[derive(Eq)]` (and a few related traits) to `EcDhSecretAgreement` then you'll get the ability back to compare them for equality.

rust-symcrypt/src/ecdh.rs
```diff
@@ -0,0 +1,257 @@
...
+
+/// EcDhKey is a wrapper around symcrypt_sys::PSYMCRYPT_ECKEY, this is to let rust handle the free'ing of this pointer
+/// EcDhKey must be allocated after EcDhCurve and free'd before EcDhCurve is free'd
+pub struct EcDhKey {
+    inner: symcrypt_sys::PSYMCRYPT_ECKEY,
+    curve: EcCurve,
+}
+
+#[derive(Debug)]
+pub struct EcDhSecretAgreement(pub Vec<u8>);
```
eholk: You may not want to make the `Vec<u8>` field `pub`, since doing so makes it part of your public API and you won't be able to change it without making a breaking change in the future. It's probably better to add something like a `.as_bytes()` method on `EcDhSecretAgreement` which returns `&[u8]` if users do need access to the bytes.

rust-symcrypt/src/ecurve.rs
mlindgren: Confirmed that the code is written this way for TLS interop. The spec for 25519 defines it to use LE. Could you just add a comment indicating that?

PhilHallin (approved):
:shipit:

PR #23, opened 2023-09-12
Updating Hashing and Hmac functions to accomodate Rustls traits 

nnmkhang:
Changes:
- added SymCryptShaXXXExpandedKey; a new wrapper for SYMCRYPT_HMAC_SHAXXX_EXPANDED_KEY
- modified HmacShaXXXInnner to ref count the expanded key 
- added copy() function in order to match rustls requirments
- added related unit tests


rust-symcrypt/src/hash.rs
```diff
@@ -1,59 +1,89 @@
...
+            state: Box::pin(Sha256InnerState(symcrypt_sys::SYMCRYPT_SHA256_STATE::default())),
+        };
+        unsafe { // SAFETY: FFI calls
+            symcrypt_sys::SymCryptSha256StateCopy(&self.state.0, &mut new_state.state.0);
+        }
+        Box::new(new_state)
+    }
 }
 
 impl Drop for Sha256State {
```
PhilHallin, line 92: >Sha256State [](http://example.com/codeflow?start=14&length=11)

Why the drop for Sha256State and not Sha256InnerState? #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,68 +1,93 @@
...
 pub const SHA256_HMAC_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA256_RESULT_SIZE as usize;
 pub const SHA384_HMAC_RESULT_SIZE: usize = symcrypt_sys::SYMCRYPT_SHA384_RESULT_SIZE as usize;
 
-pub struct HmacSha256State {
+/// Generic trait for HMACs
+pub trait Hmac {
+    type Result;
+
+    fn append(&mut self, data: &[u8]);
+    fn result(&mut self) -> Self::Result;
```
PhilHallin: > fn result(&mut self) -> Self::Result; [](http://example.com/codeflow?start=3&length=38)

Shouldn't this be:

 fn result(self) -> Self::Result; #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,68 +1,93 @@
...
                 data.as_ptr(),
                 data.len() as symcrypt_sys::SIZE_T,
             )
         }
     }
 
-    pub fn result(&mut self, result: &mut [u8; SHA256_HMAC_RESULT_SIZE]) {
-        unsafe {
-            symcrypt_sys::SymCryptHmacSha256Result(&mut self.state, result.as_mut_ptr());
+    fn result(&mut self) -> Self::Result {
```
PhilHallin: >fn result(&mut self) -> Self::Result [](http://example.com/codeflow?start=4&length=36)

fn result(mut self) -> Self::Result

No &mut #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,68 +1,93 @@
...
-            state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE::default(),
-            expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default(),
+            inner: Box::pin(HmacSha256Inner { 
+                state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE::default(),
+                expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default() 
+            }),
         };
-        unsafe {
+        unsafe { // SAFETY: FFI calls
             symcrypt_sys::SymCryptHmacSha256ExpandKey(
```
PhilHallin: This function returns SYMCRYPT_ERROR #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,68 +1,93 @@
...
+    fn result(&mut self) -> Self::Result {
+        let mut result = [0u8; SHA256_HMAC_RESULT_SIZE];
+        unsafe { // SAFETY: FFI calls
+            symcrypt_sys::SymCryptHmacSha256Result(&mut self.inner.state, result.as_mut_ptr());
         }
+        result
     }
 }
 
 impl Drop for HmacSha256State {
```
PhilHallin, line 150: >HmacSha256State [](http://example.com/codeflow?start=14&length=15)

Why not HmacSha256Inner? #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,68 +1,93 @@
...
             );
         }
     }
 }
 
 pub fn hmac_sha256(key: &[u8], data: &[u8], result: &mut [u8; SHA256_HMAC_RESULT_SIZE]) {
-    unsafe {
+    unsafe { // SAFETY: FFI calls
         let mut expanded_key = symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default();
         symcrypt_sys::SymCryptHmacSha256ExpandKey(
```
PhilHallin: >SymCryptHmacSha256ExpandKey [](http://example.com/codeflow?start=22&length=27)

Can fail. #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,68 +1,93 @@
...
-                ptr::addr_of_mut!(self.expanded_key) as *mut c_void,
-                mem::size_of_val(&mut self.expanded_key) as symcrypt_sys::SIZE_T,
+                ptr::addr_of_mut!(self.inner.expanded_key) as *mut c_void,
+                mem::size_of_val(&mut self.inner.expanded_key) as symcrypt_sys::SIZE_T,
             );
         }
     }
 }
 
 pub fn hmac_sha256(key: &[u8], data: &[u8], result: &mut [u8; SHA256_HMAC_RESULT_SIZE]) {
```
PhilHallin: >hmac_sha256 [](http://example.com/codeflow?start=7&length=11)

Can fail #Closed

rust-symcrypt/src/hash.rs
nnmkhang: I think that its cleaner, Sha256State is really just a wrapper for the Pin, it follows what we are doing in GCM as well

PhilHallin: Recommending getting the Drop closer to the inner object that could be included in other objects.

Rust ensures that drop is called for all inner objects.

rust-symcrypt/src/hmac.rs
nnmkhang: removed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,318 @@
...
+impl ExpandedKey for HmacSha256ExpandedKey {
+    type Result = symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY;
+
+    fn new() -> Self {
+        HmacSha256ExpandedKey(symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default())
+    }
+
+    fn expand(&mut self, key: &[u8]) -> Self::Result {
+        unsafe {
+            symcrypt_sys::SymCryptHmacSha256ExpandKey(
```
PhilHallin, line 39: >SymCryptHmacSha256ExpandKey [](http://example.com/codeflow?start=26&length=27)

This function can fail

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,318 @@
...
+                &mut self.0,
+                key.as_ptr(),
+                key.len() as symcrypt_sys::SIZE_T,
+            );
+        }
+        self.0
+    }
+}
```
PhilHallin, line 47: Looks like this can be removed?

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,318 @@
...
+            inner: Box::pin(HmacSha256Inner { 
+                state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE::default(),
+                expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default() 
+            }),
+        };
+        unsafe { // SAFETY: FFI calls
+            symcrypt_sys::SymCryptHmacSha256StateCopy(&self.inner.state, &self.inner.expanded_key, &mut new_state.inner.state);
         }
+        Box::new(new_state)
     }
```
PhilHallin, line 125: This won't work. The copy points back to the original expanded key.

Will need to do on your Expand Key object instead.

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,318 @@
...
+    
+    fn copy(&self) -> Box<Self> {
+        let mut new_state = HmacSha256State {
+            inner: Box::pin(HmacSha256Inner { 
+                state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE::default(),
+                expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default() 
+            }),
+        };
+        unsafe { // SAFETY: FFI calls
+            symcrypt_sys::SymCryptHmacSha256StateCopy(&self.inner.state, &self.inner.expanded_key, &mut new_state.inner.state);
```
PhilHallin: >&self.inner.expanded_key [](http://example.com/codeflow?start=73&length=24)

This expanded key will need to be ref counted. #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,318 @@
...
+    }
+}
+
 pub struct HmacSha256State {
+    inner: Pin<Box<HmacSha256Inner>>,
+}
+
+struct HmacSha256Inner {
     state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE,
     expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY,
```
PhilHallin: >symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY, [](http://example.com/codeflow?start=18&length=48)

This will need to Rc<> or Arc<> 
 For when used in the copy(). #Closed

PhilHallin: Also changed to heap allocaation pointer.

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,419 @@
...
+pub trait HmacState {
+    type Result;
+
+    fn append(&mut self, data: &[u8]);
+    fn result(self) -> Self::Result;
+    fn copy(&self) -> Box<Self>;
+}
+
+/// Wrapper for the expanded key that is Pin<Box<>>'d because we want to ensure that its memory address is heap allocated and not moved through its lifetime.
+struct HmacSha256ExpandedKey(Pin<Box<symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY>>);
```
nnmkhang: @eholk  is `Pin<Box<>>` overkill here? Im assuming `Arc<>` must have some sort of built in Boxing? I think we'd rather be safe than sorry here with SymCrypt's memory managment stuff #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,419 @@
...
+/// SymCrypt expects the address for its structs to stay static through the structs lifetime to guarantee that structs are not memcpy'd as
+/// doing so would lead to use-after-free and inconsistent states.
 pub struct HmacSha256State {
+    inner: Pin<Box<HmacSha256Inner>>,
+}
+
+struct HmacSha256Inner {
     state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE,
-    expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY,
+    expanded_key: Arc<HmacSha256ExpandedKey>,
```
nnmkhang: @mlindgren just FYI, we are using `Arc<>` to ref count the expanded key since `SymCryptHmacShaXXXCopy()` takes reference to the original expanded key

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,419 @@
...
+}
+
+unsafe impl Send for HmacSha256Inner {
```
nnmkhang, line 65: Based on the rustls `Send + Sync` parameters on their traits. We will follow up with them offline to see their requirements.  I think that async can get a little dicey here so we'd like some more context before implementing.  #Closed

nnmkhang (commented):
LGTM ( i wrote it )

rust-symcrypt/src/hmac.rs
PhilHallin: As a minimum it must be Box'ed so it is allocated on the heap with a pointer to that doesn't change to satisfy symcrypt magic requirments.

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,419 @@
...
 }
 
 impl HmacSha256State {
-    pub fn new(key: &[u8]) -> Self {
-        let mut instance = HmacSha256State {
-            state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE::default(),
-            expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default(),
-        };
+    pub fn new(key: &[u8]) -> Result<Self, SymCryptError> {
+        let mut expanded_key = Pin::new(Box::new(
```
PhilHallin: >expanded_key = [](http://example.com/codeflow?start=16&length=14)

You might want this to be HmacSha256ExpandedKey(Pin::new(Box::new( ...

So that your drop would be called to wipe for any errors.

This will be the same pattern for the stateless call that needs to do an expand key. #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,419 @@
...
                 key.as_ptr(),
                 key.len() as symcrypt_sys::SIZE_T,
-            );
-            symcrypt_sys::SymCryptHmacSha256Init(&mut instance.state, &instance.expanded_key);
+            ) {
+                symcrypt_sys::SYMCRYPT_ERROR_SYMCRYPT_NO_ERROR => {
+                    let mut instance = HmacSha256State {
+                        inner: Box::pin(HmacSha256Inner {
+                            state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE::default(),
+                            expanded_key: Arc::new(HmacSha256ExpandedKey(expanded_key)),
```
PhilHallin: >HmacSha256ExpandedKey [](http://example.com/codeflow?start=51&length=21)

Do this above. #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,419 @@
...
-pub fn hmac_sha256(key: &[u8], data: &[u8], result: &mut [u8; SHA256_HMAC_RESULT_SIZE]) {
+/// Stateless Hmac functions for Sha256
+pub fn hmac_sha256(
+    key: &[u8],
+    data: &[u8],
+    result: &mut [u8; SHA256_HMAC_RESULT_SIZE],
+) -> Result<(), SymCryptError> {
     unsafe {
+        // SAFETY: FFI calls
         let mut expanded_key = symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default();
```
PhilHallin: >  let mut expanded_key [](http://example.com/codeflow?start=6&length=22)

You want to wrap this via HmacSha256ExpandedKey() so it naturally gets dropped and wiped. #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,419 @@
...
+                symcrypt_sys::SymCryptHmacSha256(
+                    &mut expanded_key,
+                    data.as_ptr(),
+                    data.len() as symcrypt_sys::SIZE_T,
+                    result.as_mut_ptr(),
+                );
+
+                symcrypt_sys::SymCryptWipe(
+                    ptr::addr_of_mut!(expanded_key) as *mut c_void,
+                    mem::size_of_val(&mut expanded_key) as symcrypt_sys::SIZE_T,
```
PhilHallin: This can be removed when you wrap via HmacSha256ExpandedKey #Closed

rust-symcrypt/src/hmac.rs
PhilHallin: You might want to double check that drop isn't called here for the success case?

nnmkhang: Changed

rust-symcrypt/src/hmac.rs
nnmkhang: Changed

rust-symcrypt/src/hmac.rs
nnmkhang: Removed

rust-symcrypt/src/hash.rs
```diff
@@ -1,59 +1,98 @@
+/* Hashing functions. For further documentation please refer to symcrypt.h */
```
eholk:
```suggestion
//! Hashing functions. For further documentation please refer to symcrypt.h
```

This makes it a module-level doc comment, so it will show up in docs.rs.

rust-symcrypt/src/hash.rs
```diff
@@ -1,59 +1,98 @@
...
 
-pub trait Hash {
+/// Generic trait for stateful hashing
+///
+/// append() appends to be hashed data to the state, this operation can be done multiple times.
+///
+/// result() returns the result of the hash. Once result() is called, the lifetime of the ShaXXXState is finished.
+/// To perform other stateful hash operation you must create a new hash object via ShaXXXState::new()
+///
+/// copy() creates a copy of the current ShaXXXState
```
eholk:
```suggestion
/// Generic trait for stateful hashing
///
/// `append()` appends to be hashed data to the state, this operation can be done multiple times.
///
/// `result()` returns the result of the hash. Once `result()` is called, the lifetime of the ShaXXXState is finished.
/// To perform other stateful hash operation you must create a new hash object via ShaXXXState::new()
///
/// `copy()` creates a copy of the current ShaXXXState
```

Just adding some markdown formatting for the rustdoc pages. I think you can also write them as ``[`append`]`` and then it will also link to the docs for the `append` function.

rust-symcrypt/src/hash.rs
```diff
@@ -1,59 +1,98 @@
...
+pub trait HashState {
     type Result;
 
     fn append(&mut self, data: &[u8]);
-    fn result(&mut self) -> Self::Result;
+    fn result(self) -> Self::Result;
+    fn copy(&self) -> Box<Self>;
```
eholk: From a Rust standpoint, the `copy` function strikes me as a little weird. In Rust a copy is usually a bitwise copy (if your type impls the `Copy` trait that means the compiler can safely duplicate it with `memcpy`), and `clone()` is used for more complex copies. That said, I wouldn't just rename `copy` to `clone`, since `clone` is expected to come from the `Clone` trait instead.

One option is to redeclare `HashState` as `pub trait HashState: Clone`, which means everything that implements `HashState` can also be cloned. The problem is the way you've written this now the copy is always in a box and we wouldn't be able to guarantee that with `pub trait HashState: Clone`.

Another option is to have the `HashXXXState`s all `impl Clone for Box<HashXXXState>`, which would let you clone a box of a hash state and guarantee the copy is in a box.

Anyway, it's probably better for now to leave things mostly like you have them. Maybe rename it to `duplicate` or something to avoid the semantic clash with `copy`. I'll leave this up to your judgment though.  #Closed

rust-symcrypt/src/hash.rs
```diff
@@ -1,59 +1,98 @@
...
 }
 
+/// Sha256State needs to have a heap allocated inner state that is Pin<Box<T>> since the memory address of Self is moved around when implementing
+/// HashState Result field.
+///
+/// SymCrypt expects the address for its structs to stay static through the structs lifetime to guarantee that structs are not memcpy'd as
+/// doing so would lead to use-after-free and inconsistent states.
 pub struct Sha256State {
```
eholk, line 30:
```suggestion
#[derive(Default)]
pub struct Sha256State {
```

rust-symcrypt/src/hash.rs
```diff
@@ -1,59 +1,98 @@
...
+struct Sha256InnerState(symcrypt_sys::SYMCRYPT_SHA256_STATE);
+
+/// Creates a new instance of Sha256State, this must be called before other HashState functions can be called
 impl Sha256State {
     pub fn new() -> Self {
         let mut instance = Sha256State {
-            state: symcrypt_sys::SYMCRYPT_SHA256_STATE::default(),
+            state: Box::pin(Sha256InnerState(
+                symcrypt_sys::SYMCRYPT_SHA256_STATE::default(),
+            )),
```
eholk, line 42:
```suggestion
            state: Box::pin(Sha256InnerState::default()),
```

If you stick `#[derive(Default)]` on the struct definition you can make this a little simpler.

rust-symcrypt/src/hash.rs
```diff
@@ -1,59 +1,98 @@
...
-                &mut self.state,
+                &mut self.state.0,
                 data.as_ptr(),
                 data.len() as symcrypt_sys::SIZE_T,
             );
         }
     }
 
-    fn result(&mut self) -> Self::Result {
+    fn result(mut self) -> Self::Result {
```
eholk: Are you supposed to be able to reuse this object after getting the result?

rust-symcrypt/src/hash.rs
```diff
@@ -1,59 +1,98 @@
...
+        let mut new_state = Sha256State {
+            state: Box::pin(Sha256InnerState(
+                symcrypt_sys::SYMCRYPT_SHA256_STATE::default(),
+            )),
+        };
+        unsafe {
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptSha256StateCopy(&self.state.0, &mut new_state.state.0);
+        }
+        Box::new(new_state)
```
eholk: Oh wait! I don't think this box is needed, since `Sha256State` is just a pointer to the inner state which stays pinned in memory. Putting that inside a box just makes a double indirection.

This also means I can revise my earlier comment about `copy`, so I'm going to go reply to my previous one.

rust-symcrypt/src/hash.rs
eholk: After seeing how `copy` is implemented and how the various `ShaXXXState` structs work, I think I would suggest just using `Clone` instead. You don't need to keep `Self` in a box here, because your structs all add in internal indirection to keep the pinning guarantees that are needed. This means the clone impl can create any new memory allocations and such that are needed. Let me know if that makes sense!

rust-symcrypt/src/hash.rs
```diff
@@ -74,47 +115,74 @@ pub fn sha256(data: &[u8]) -> [u8; SHA256_RESULT_SIZE] {
...
 }
 
+/// Creates a new instance of Sha384State, this must be called before other HashState functions can be called
 impl Sha384State {
     pub fn new() -> Self {
         let mut instance = Sha384State {
-            state: symcrypt_sys::SYMCRYPT_SHA384_STATE::default(),
+            state: Box::pin(Sha384InnerState(
+                symcrypt_sys::SYMCRYPT_SHA384_STATE::default(),
+            )),
```
eholk, line 135: The `#[derive(Default)]` trick would work here too.

rust-symcrypt/src/hash.rs
```diff
@@ -139,7 +209,7 @@ pub fn sha384(data: &[u8]) -> [u8; SHA384_RESULT_SIZE] {
 mod test {
```
eholk, line 212: I like the way you've added tests here with the generic test methods and then specific ones for each type!

rust-symcrypt/src/hmac.rs
eholk: The `Pin<Box<...>>` is overkill here. Instead, I'd change line 60 to say `expanded_key: Pin<Arc<HmacSha256ExpandedKey>>` (I'll leave a comment there too).

Both `Box` and `Arc` are heap allocations. The difference is that `Box` is uniquely owned, while `Arc` can have shared ownership. `Box` is like `unique_ptr` in C++ and `Arc` is like `shared_ptr`.

rust-symcrypt/src/hmac.rs
```diff
@@ -1,180 +1,409 @@
...
+/// SymCrypt expects the address for its structs to stay static through the structs lifetime to guarantee that structs are not memcpy'd as
+/// doing so would lead to use-after-free and inconsistent states.
 pub struct HmacSha256State {
+    inner: Pin<Box<HmacSha256Inner>>,
+}
+
+struct HmacSha256Inner {
     state: symcrypt_sys::SYMCRYPT_HMAC_SHA256_STATE,
-    expanded_key: symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY,
+    expanded_key: Arc<HmacSha256ExpandedKey>,
```
eholk:
```suggestion
    expanded_key: Pin<Arc<HmacSha256ExpandedKey>>,
```

If we pin the `Arc` here then we don't have to have an additional `Pin<Box<...>>` (and an extra level of indirection) in the expanded key struct. #Closed

rust-symcrypt/src/hmac.rs
eholk: For this impl to be safe, you'll need to make sure SymCrypt itself is thread-safe. Alternatively, you can wrap the `HmacSha256Inner` field in `HmacSha256State` with a `Mutex` and then `HmacSha256State` should automatically be `Send` without needing any `unsafe impl`s.

rust-symcrypt/src/hash.rs
nnmkhang: Noted, I will change for other files as well

rust-symcrypt/src/hash.rs
nnmkhang: noted, will change to ['append`]

rust-symcrypt/src/hmac.rs
PhilHallin: However, this needs to remain a Box so it remains as a heap pointer.

rust-symcrypt/src/hmac.rs
PhilHallin: Let's first talk to the rustls folks before adding a Mutex.

I'm not seeing other implementations like ring adding a Mutex.

nnmkhang: I think for now we will table the `Send + Sync` changes until we get a better idea of what is expected from the rustls side in terms of Async.

rust-symcrypt/src/hmac.rs
nnmkhang: The reason we have `Pin<Box<>>` here is because in the `new()` we first initialize the variable `expanded_key` on line 73 and we are worried that rust might move the memory address before we are able to `Pin<Arc<>>` on line 86,

nnmkhang: I will try to remove the `Pin<Box<>>` and see if there is any errors in the unit test

rust-symcrypt/src/hash.rs
PhilHallin: Correct for symcrypt. However, in the rustls wrapper it doesn't support reuse.

rust-symcrypt/src/hash.rs
PhilHallin: I see that rustls used fork for this function name.

rust-symcrypt/src/hash.rs
nnmkhang: Yes we are, we made a mistake. Changing back to be able to be used multiple times

rust-symcrypt/src/hash.rs
nnmkhang: I thought that this would work easily but looks like Rust cant pick up the `#[derive(Default)]` for `symcrypt_sys::SYMCRYPT_SHA256_STATE`, error : 

> the trait bound `Pin<Box<Sha256InnerState>>: Default` is not satisfied
> the trait `Default` is not implemented for `Pin<Box<Sha256InnerState>>`

Even after I added `#[derive(Default)]` for both `Sha256State` and `Sha256InnerState`

for now I will keep how I have it

rust-symcrypt/src/hash.rs
nnmkhang: See other comment on `#[derive(Default)]`

rust-symcrypt/src/hash.rs
nnmkhang: thanks :)

rust-symcrypt/src/hash.rs
nnmkhang: > After seeing how `copy` is implemented and how the various `ShaXXXState` structs work, I think I would suggest just using `Clone` instead. You don't need to keep `Self` in a box here, because your structs all add in internal indirection to keep the pinning guarantees that are needed. This means the clone impl can create any new memory allocations and such that are needed. Let me know if that makes sense!

This makes sense, I have moved the copy over to.

`impl Clone for ShaXXXState` 

Unit tests work so I think it should be fine.

rust-symcrypt/src/hash.rs
eholk: 👍 

After suggesting that I was playing around a little and realized `Pin` doesn't impl `Default`, so the suggestion doesn't quite work :)

rust-symcrypt/src/hmac.rs
```diff
@@ -1,190 +1,436 @@
...
+                ptr::addr_of_mut!(self.0) as *mut c_void,
+                mem::size_of_val(&mut self.0) as symcrypt_sys::SIZE_T,
+            );
+        }
+    }
+}
+
+/// Using an inner HmacSha256 state that is Pin<Box<>>'d since the memory address for Self is moved around when returning from HmacSha256State::new()
+///
+/// expanded_key is Arc<>'d to be properly ref counted when calling HmacShaXXXState::copy().
```
PhilHallin: >copy [](http://example.com/codeflow?start=85&length=4)

clone() #Closed

nnmkhang: changed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,190 +1,436 @@
...
+    data: &[u8],
+) -> Result<[u8; SHA256_HMAC_RESULT_SIZE], SymCryptError> {
+    let mut result = [0u8; SHA256_HMAC_RESULT_SIZE];
     unsafe {
-        let mut expanded_key = symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default();
-        symcrypt_sys::SymCryptHmacSha256ExpandKey(
-            &mut expanded_key,
+        // SAFETY: FFI calls
+        let mut expanded_key = Arc::new(HmacSha256ExpandedKey(
+            symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default(),
```
PhilHallin: This expanded_key won't  need to be shared. Don't need the Arc::new()

= HmacSha256ExpandedKey(symcrypt_sys::SYMCRYPT_HMAC_SHA256_EXPANDED_KEY::default(),); #Closed

rust-symcrypt/src/hmac.rs
```diff
@@ -1,190 +1,436 @@
...
+    key: &[u8],
+    data: &[u8],
+) -> Result<[u8; SHA384_HMAC_RESULT_SIZE], SymCryptError> {
+    let mut result = [0u8; SHA384_HMAC_RESULT_SIZE];
     unsafe {
-        let mut expanded_key = symcrypt_sys::SYMCRYPT_HMAC_SHA384_EXPANDED_KEY::default();
-        symcrypt_sys::SymCryptHmacSha384ExpandKey(
-            &mut expanded_key,
+        // SAFETY: FFI calls
+        let mut expanded_key = Arc::new(HmacSha384ExpandedKey(
```
PhilHallin: >Arc::new [](http://example.com/codeflow?start=31&length=8)

Won't be shared. Don't need the Arc:::new() #Closed

PhilHallin (approved):
:shipit:

PR #25, opened 2023-09-28
User/nnmkhang/update ecdh for rustls

nnmkhang:
Changed:
- Renamed eccurve.rs to eckey 
- Moved over key functionality to eckey
- Updated comments for EcDh, Hmac and Hashing
- Added display fmt for SymCryptError

Todo:
- Discuss Send+Sync traits for bindings
- research lazy static


rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,156 @@
...
+///
+/// [`new()`] returns a new EcKey object that has the key and curve allocated. This key must be allocated after the [`EcCurve`] has been allocated
+/// and dropped after [`EcCurve`] has been dropped.
+///
+/// [`inner()`] is an accessor to the inner field of the EcKey struct. Reference is not needed here since we are working with a raw SymCrypt pointer.
+///
+/// [`curve()`] is an accessor to the curve field of the EcKey struct. Reference is used here since EcKey should still maintain ownership of the EcCurve.
+impl EcKey {
+    pub(crate) fn new(curve: CurveType) -> Result<Self, SymCryptError> {
+        let ecdh_curve = EcCurve::new(curve)?;
```
PhilHallin, line 40: >  let ecdh_curve = EcCurve::new(curve)?; [](http://example.com/codeflow?start=6&length=40)

Since you return an error for null key_ptr, might also want to return error for curve failure.

rust-symcrypt/src/ecdh.rs
```diff
@@ -21,51 +15,33 @@ impl EcDhSecretAgreement {
...
+///
+/// ['from_public_key_bytes()'] takes in a public_key and creates a EcDh struct who's EcKey has only a public key attached.
+///
+/// [`get_public_key_bytes()`] returns a Vec<u8> that is the public key associated with the current EcKey
+///  
+/// [`ecdh_secret_agreement()`] takes in two EcDh structs and returns the associated secret agreement.
 impl EcDh {
-    /// EcDh::new() returns an EcDh struct that has a private/public key pair.
     pub fn new(curve: CurveType) -> Result<Self, SymCryptError> {
+        let ecdh_key = EcKey::new(curve)?;
```
PhilHallin, line 39: > let ecdh_key = EcKey::new(curve)?; [](http://example.com/codeflow?start=7&length=35)

If this fails, shouldn't it return an error?

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        unsafe {
+            // SAFETY: FFI calls
+            let key_ptr = symcrypt_sys::SymCryptEckeyAllocate(ecdh_curve.0); // stack allocated since will do SymCryptEckeyAllocate.
```
PhilHallin, line 44: >stack [](http://example.com/codeflow?start=80&length=5)

heap?

rust-symcrypt/src/eckey.rs
PhilHallin: >ecdh_curve  [](http://example.com/codeflow?start=12&length=11)

ec_curve #Closed

rust-symcrypt/src/ecdh.rs
```diff
@@ -177,8 +151,8 @@ mod test {
     #[test]
     fn test_ecdh_nist_p256() {
         symcrypt_init(); // must run symcrypt_init for the alloc callbacks to initialize
```
PhilHallin: > symcrypt_init() [](http://example.com/codeflow?start=7&length=16)

Why doesn't EcDh::new() call symcrypt_init? #Closed

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+}
+
+/// Impl for EcCurve
+///
+/// [`new()`] returns a [`EcCurve`] associated with the provided [`CurveType`].
+///
+/// [`get_size`] returns the size of the [`EcCurve`] as a u32.
+impl EcCurve {
+    pub(crate) fn new(curve: CurveType) -> Result<Self, SymCryptError> {
```
PhilHallin: >new [](http://example.com/codeflow?start=18&length=3)

This function should call symcrypt_init() #Closed

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+/// Impl for EcCurve
+///
+/// [`new()`] returns a [`EcCurve`] associated with the provided [`CurveType`].
+///
+/// [`get_size`] returns the size of the [`EcCurve`] as a u32.
+impl EcCurve {
+    pub(crate) fn new(curve: CurveType) -> Result<&'static Self, SymCryptError> {
```
PhilHallin: >Result<&'static Self, SymCryptError> { [](http://example.com/codeflow?start=43&length=38)

This won't need the Result wrapper anymore. #Closed

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,167 @@
...
+impl Drop for EcKey {
+    fn drop(&mut self) {
+        unsafe {
+            // SAFETY: FFI calls
+            symcrypt_sys::SymCryptEckeyFree(self.inner);
+        }
+    }
+}
+
+// Curves can be re-used across EcDh calls, creating static references to save on allocations and increase perf.
```
PhilHallin: >EcDh [](http://example.com/codeflow?start=32&length=4)

EcKey #Closed

rust-symcrypt/src/ecdh.rs
```diff
@@ -1,17 +1,11 @@
...
-/// EcDhKey is a wrapper around symcrypt_sys::PSYMCRYPT_ECKEY, this is to let rust handle the free'ing of this pointer
-/// EcDhKey must be allocated after EcDhCurve and free'd before EcDhCurve is free'd
-pub struct EcDhKey {
-    inner: symcrypt_sys::PSYMCRYPT_ECKEY,
-    curve: EcCurve,
-}
-
+/// Wrapper for the EcDh secret agreement result value. This is in place to make the return clear to the caller.
 #[derive(Debug)]
 pub struct EcDhSecretAgreement(Vec<u8>);
```
eholk, line 10: On option to consider here is using `Box<[u8]>` instead of `Vec<u8>`. `Box<[u8]>` works if you don't need to dynamically resize the array after it's created and it saves you a word for the capacity of the buffer.

I wouldn't worry about that for this PR, but it's a possibility for future cleanup.

rust-symcrypt/Cargo.toml
```diff
@@ -8,6 +8,7 @@ edition = "2021"
 [dependencies]
 symcrypt-sys = {path = "../symcrypt-sys"}
 libc = "0.2.0"
+lazy_static = "1.4.0"
```
eholk, line 11: I saw you had a TODO to research lazy_static. You might also consider `OnceCell` or [`OnceLock`](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) since these are built into the standard library now.

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,167 @@
+//! Friendly rust types for CurveTypes.
+
+use lazy_static::lazy_static;
+use crate::{errors::SymCryptError, symcrypt_init};
+use symcrypt_sys;
+
+/// [`CurveType`] provides an enum of the curve types that can be used when creating an [`EcKey`].
+#[derive(Copy, Clone, PartialEq, Debug)]
+pub enum CurveType {
```
eholk, line 9: Is it expected that there might be new curve types in the future? If so, it'd probably be good to add `#[non_exhaustive]` to the enum. That requires people to add a catch-all clause when matching on this so that if you add variants in the future it won't cause a compiler error.

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/// [`EcKey`] is a wrapper around symcrypt_sys::PSYMCRYPT_ECKEY, this is to let rust handle the drop and subsequent free.
+/// EcKey must be allocated after [`EcCurve`] and free'd before EcCurve is free'd.
+///
+/// Allocation for EcKey is handled by SymCrypt via SymCryptEcKeyAllocate, and is subsequently stored on the stack, therefore pointer will
+/// not move and Box<> is not needed.
+pub struct EcKey {
+    inner: symcrypt_sys::PSYMCRYPT_ECKEY,
+    curve: &'static EcCurve,
```
eholk, line 20: I'm not sure `'static` is quite right here, at least based on the comment. `'static` means the value pointed to will be live forever, but the documentation says the `EcKey` needs to be freed before the `EcCurve` is freed, which should never happen if you have a true `'static` reference to it. #Closed

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,167 @@
...
+    NistP256,
+    NistP384,
+    Curve25519,
+}
+
+/// [`EcKey`] is a wrapper around symcrypt_sys::PSYMCRYPT_ECKEY, this is to let rust handle the drop and subsequent free.
+/// EcKey must be allocated after [`EcCurve`] and free'd before EcCurve is free'd.
+///
+/// Allocation for EcKey is handled by SymCrypt via SymCryptEcKeyAllocate, and is subsequently stored on the stack, therefore pointer will
+/// not move and Box<> is not needed.
```
eholk: I'd move this comment to the `inner` field. Where it is now means it will show up in the public API documentation, and users of this library shouldn't need to care about whether a `Box` is needed on a field they can't even see. Documenting this rationale is valuable to future editors of this code though!

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,167 @@
...
+///
+/// [`new()`] returns a new EcKey object that has the key and curve allocated. This key must be allocated after the [`EcCurve`] has been allocated
+/// and dropped after [`EcCurve`] has been dropped.
+///
+/// [`inner()`] is an accessor to the inner field of the EcKey struct. Reference is not needed here since we are working with a raw SymCrypt pointer.
+///
+/// [`curve()`] is an accessor to the curve field of the EcKey struct. Reference is used here since EcKey should still maintain ownership of the EcCurve.
+impl EcKey {
+    pub(crate) fn new(curve: CurveType) -> Result<Self, SymCryptError> {
+        let ec_curve = &EcCurve::new(curve);
```
eholk:
```suggestion
        let ec_curve = EcCurve::new(curve);
```

I don't think you need the `&` here.

rust-symcrypt/src/eckey.rs
```diff
@@ -0,0 +1,167 @@
...
+
+unsafe impl Send for EcCurve {
+    // TODO: Figure out error :
+    // *const _SYMCRYPT_ECURVE_PARAMS cannot be shared between threads safely
+    // the trait Sync is not implemented for *const _SYMCRYPT_ECURVE_PARAMS
+}
+
+unsafe impl Sync for EcCurve {
+    // ??
+}
```
eholk, line 79: Usually you don't need to implement these yourself, but this might be a case where that's needed (or at least reasonable). The key thing is whether the underlying SymCrypt APIs are thread-safe (meaning you can access things like `SYMCRYPT_ECURVE_PARAMS` from multiple threads at the same time.

Most things should be `Send`, meaning you can transfer ownership to a different thread. In that case, once you send a value, you can't access it from the previous thread anymore, so there's no risk of data races.

`Sync` means you can refer to the _same object_ from multiple threads simultaneously. Usually this means it has some kind of internal locking to make it safe.

Rust can generally figure out what things are `Send` and `Sync`, but it's conservative around raw pointers (e.g. `*const`) because these do not have as strong of invariants around them. When you implement `Send` and `Sync`, it means you're promising that these types meet all the requirements to be sendable and accessed from multiple threads.

Since these are `unsafe` to implement, I'd strongly recommend adding a `// SAFETY: ` comment explaining why these are safe to share between threads or transfer between threads.

eholk (approved):

rust-symcrypt/src/eckey.rs
PhilHallin: Comment wasn't correct/

PhilHallin (approved):
:shipit:

PR #26, opened 2023-11-02
Removing unneeded Sha256InnerState

nnmkhang:
Removing unneeded Inner struct, initilizting via `let mut instance = Sha256State(Box::pin(symcrypt_sys::SYMCRYPT_SHA256_STATE::default()));` does the trick, Inner struct is not needed.


PhilHallin (approved):
:shipit:

